PR #70, opened 2026-07-07
RS-68: Enforce request-body validation with @Valid and an OnUpdate group

UnLow1:
## Ticket

[RS-68 — Walidacja DTO jest martwa — dodać @Valid albo usunąć starter](https://referee-staffer.youtrack.cloud/issue/RS-68)

The DTOs carry `@NotNull` annotations, but no controller ever used `@Valid`/`@Validated`, so bean validation never ran. Requests with missing fields reached the services and either failed deeper (as 500s) or persisted nulls. The ticket offered two options: wire the validation up, or remove `spring-boot-starter-validation`. This PR implements option 1 — the starter is already on the classpath, and the frontend already renders `ProblemDetail.detail` in its global error toast, so enabled validation becomes user-visible for free.

## Plan

1. **Confirm the create/update payload split.** The form drawers build create payloads *without* an id (the server generates it), while every update flow sends a full object including id — and the update code paths dereference `dto.getId()`. Blindly activating `@NotNull id` would break every create.
2. **Introduce an `OnUpdate` validation group** (`extends Default`) and move the id constraint into it: POST validates with plain `@Valid` (Default group, id optional), PUT with `@Validated(OnUpdate.class)` (Default + id required).
3. **Review DTO annotations for completeness**: `@NotBlank` for strings, `@NotNull` for `IDRequest.ids`, and constraints on `Config` (which doubles as the `PUT /api/configuration` request body).
4. **Handle list bodies** (`PUT /api/matches`, `PUT /api/configuration`) via `List<@Valid Dto>` element validation.
5. **Map validation failures in `RestExceptionHandler`** to a ProblemDetail 400 whose `detail` lists the offending fields, matching the handler's existing style.
6. **Cover with `@WebMvcTest` specs**: 400 features per controller, plus fixing existing specs whose request bodies become invalid once validation is on.

## Changes

- **`model/validation/OnUpdate.java`** (new) — marker group extending `Default`; an update payload must satisfy all create-time rules plus a non-null id.
- **DTOs** — `id` is now `@NotNull(groups = OnUpdate.class)` in Grade/Match/Referee/Team/Vacation DTOs; string fields tightened from `@NotNull` to `@NotBlank` (`RefereeDto.firstName/lastName/email`, `TeamDto.name/city`); `IDRequest.ids` gains `@NotNull`; `Config.name/value` gain `@NotNull` mirroring their `NOT NULL` columns so a bad configuration payload 400s at the controller instead of blowing up as a 500 on flush.
- **Controllers** — `@Valid` on all POST bodies (including the `byIds` endpoints), `@Validated(OnUpdate.class)` on all PUT bodies, `List<@Valid ...>` on the two bulk PUT endpoints.
- **`RestExceptionHandler`** — two new handlers: `MethodArgumentNotValidException` (single `@Valid` body) and `HandlerMethodValidationException` (Spring routes `List<@Valid T>` bodies through built-in method validation, which throws this instead). Both return ProblemDetail 400 with a sorted `field: message` list in `detail` — the one property the frontend toast renders. List-element errors are prefixed with their container index (`[1].queue: must not be null`).
- **`pom.xml`** — Surefire pins `-Duser.language=en -Duser.country=US` (appended to JaCoCo's `@{argLine}` so the coverage agent keeps working): the specs assert Hibernate Validator's default messages, which resolve against the JVM locale.

Design decisions worth calling out:

- **Groups over dropping the id constraint**: removing `@NotNull` from id entirely would let a PUT without id slip through to `findById(null)` / `save()`-as-insert — exactly the class of deep failure this ticket is about.
- **Bulk `PUT /api/matches` validates elements in the Default group only** — element-level group selection isn't supported by Spring's container validation, so a missing id in a bulk element is not caught by validation. Acceptable: the only caller (Staffer save flow) sends server-produced DTOs that always carry ids.
- **`updateConfiguration` keeps accepting the `Config` entity** as its body (as before); this PR only adds constraints. Splitting a dedicated DTO was out of scope.

## Testing

- `mvn -DskipFrontend=true test` — **128 tests, 0 failures** (JDK 25).
- 12 new `@WebMvcTest` feature methods across all six body-accepting controllers: 400 for incomplete create bodies, 400 for PUT without id, 400 for a malformed email, 400 for a missing `ids` list, 400 for an incomplete bulk-list element (asserting the `[index].field` prefix), each also verifying the mocked service/repository is never touched.
- Existing specs whose bodies became invalid under the new rules were updated to send complete payloads (referee create/update, team update-404, match update/bulk).
- Frontend flows were checked against the new rules: every required backend field is also required in the corresponding form drawer, and create payloads legitimately omit id — covered by the create specs, which post bodies without id and expect 200.

https://claude.ai/code/session_01N1wTd5UPeew574xMr6aPP7

---
_Generated by [Claude Code](https://claude.ai/code/session_01N1wTd5UPeew574xMr6aPP7)_

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,6 +134,10 @@
                     <includes>
                         <include>**/*Spec.java</include>
                     </includes>
+                    <!-- Controller specs assert Hibernate Validator's default constraint
+                         messages, which are resolved against the JVM locale — pin English
+                         so the suite passes on non-English machines too. -->
+                    <argLine>@{argLine} -Duser.language=en -Duser.country=US</argLine>
                 </configuration>
             </plugin>
 
--- a/src/main/java/com/jamex/refereestaffer/controller/MatchController.java
+++ b/src/main/java/com/jamex/refereestaffer/controller/MatchController.java
@@ -4,6 +4,7 @@ import com.jamex.refereestaffer.model.converter.MatchConverter;
 import com.jamex.refereestaffer.model.dto.DifficultyBreakdownDto;
 import com.jamex.refereestaffer.model.dto.MatchDto;
 import com.jamex.refereestaffer.model.exception.MatchNotFoundException;
+import com.jamex.refereestaffer.model.exception.RequestValidationException;
 import com.jamex.refereestaffer.model.validation.OnUpdate;
 import com.jamex.refereestaffer.repository.MatchRepository;
 import com.jamex.refereestaffer.service.MatchService;
@@ -20,6 +21,7 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
@@ -83,6 +85,7 @@ public class MatchController {
 
     @PutMapping
     public void updateMatches(@RequestBody List<@Valid MatchDto> matchesDtos) {
+        requireIds(matchesDtos);
         var matchIds = matchesDtos.stream()
                 .map(MatchDto::getId)
                 .toList();
@@ -101,5 +104,22 @@ public class MatchController {
     public void deleteMatch(@PathVariable Long id) {
         log.info("Deleting match with id = {}", id);
         matchService.deleteMatch(id);
+    }
+
+    /**
+     * Element validation on a {@code List<@Valid ...>} body always runs in the Default
+     * group (container validation cannot select OnUpdate), so id presence must be checked
+     * by hand — a null id would make saveAll() insert a new match instead of updating one.
+     */
+    private static void requireIds(List<MatchDto> matchesDtos) {
+        var errors = new ArrayList<String>();
+        for (int i = 0; i < matchesDtos.size(); i++) {
+            if (matchesDtos.get(i).getId() == null) {
+                errors.add("[" + i + "].id: must not be null");
+            }
+        }
+        if (!errors.isEmpty()) {
+            throw new RequestValidationException(String.join("; ", errors));
+        }
     }
 }
--- a/src/main/java/com/jamex/refereestaffer/controller/MatchController.java
+++ b/src/main/java/com/jamex/refereestaffer/controller/MatchController.java
@@ -4,10 +4,13 @@ import com.jamex.refereestaffer.model.converter.MatchConverter;
 import com.jamex.refereestaffer.model.dto.DifficultyBreakdownDto;
 import com.jamex.refereestaffer.model.dto.MatchDto;
 import com.jamex.refereestaffer.model.exception.MatchNotFoundException;
+import com.jamex.refereestaffer.model.validation.OnUpdate;
 import com.jamex.refereestaffer.repository.MatchRepository;
 import com.jamex.refereestaffer.service.MatchService;
+import jakarta.validation.Valid;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
@@ -62,7 +65,7 @@ public class MatchController {
     }
 
     @PostMapping
-    public MatchDto createMatch(@RequestBody MatchDto matchDto) {
+    public MatchDto createMatch(@Valid @RequestBody MatchDto matchDto) {
         log.info("Adding new match");
         var match = matchConverter.convertFromDto(matchDto);
         var savedMatch = matchRepository.save(match);
@@ -71,7 +74,7 @@ public class MatchController {
 
     // TODO is this id needed?
     @PutMapping("/{id}")
-    public MatchDto updateMatch(@RequestBody MatchDto matchDto, @PathVariable Long id) {
+    public MatchDto updateMatch(@Validated(OnUpdate.class) @RequestBody MatchDto matchDto, @PathVariable Long id) {
         log.info("Updating match with id {}", matchDto.getId());
         var match = matchConverter.convertFromDto(matchDto);
         var updatedMatch = matchRepository.save(match);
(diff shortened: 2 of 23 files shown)


src/main/java/com/jamex/refereestaffer/controller/MatchController.java
```diff
@@ -71,15 +74,15 @@ public MatchDto createMatch(@RequestBody MatchDto matchDto) {
...
+    public MatchDto updateMatch(@Validated(OnUpdate.class) @RequestBody MatchDto matchDto, @PathVariable Long id) {
         log.info("Updating match with id {}", matchDto.getId());
         var match = matchConverter.convertFromDto(matchDto);
         var updatedMatch = matchRepository.save(match);
         return matchConverter.convertFromEntity(updatedMatch);
     }
 
     @PutMapping
-    public void updateMatches(@RequestBody List<MatchDto> matchesDtos) {
+    public void updateMatches(@RequestBody List<@Valid MatchDto> matchesDtos) {
```
UnLow1, line 87: Known limitation: element validation on a `List<@Valid T>` body always runs in the Default group — Spring's container validation has no per-element group selection, so a bulk element with a missing `id` is *not* rejected here (it would be persisted as a new row by `saveAll`). Acceptable today because the only caller is the Staffer save flow, which round-trips server-produced DTOs that always carry ids. If this endpoint ever gets a second caller, consider an explicit null-id check that throws, or a dedicated bulk-update DTO with a plain `@NotNull id`.

---

src/main/java/com/jamex/refereestaffer/model/entity/Config.java
```diff
@@ -15,10 +16,15 @@ public class Config {
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
 
+    // Bean-validation mirrors of the NOT NULL columns — the entity doubles as the
+    // request body of PUT /api/configuration, so nulls must 400 at the controller
+    // instead of surfacing as a 500 constraint violation on flush.
+    @NotNull
```
UnLow1, line 22: Side effect worth knowing: because `Config` is a JPA entity, Hibernate also evaluates these bean-validation constraints at flush time (pre-insert/pre-update events), not just in the controller. That's harmless here — the constraints exactly mirror the `NOT NULL` columns, so anything they'd reject would have failed at the DB anyway — but it's the reason I'd stop short of adding *richer* validation (ranges, formats) to this entity. If that's ever needed, split a request DTO for `PUT /api/configuration` instead.

---

pom.xml
```diff
@@ -134,6 +134,10 @@
                     <includes>
                         <include>**/*Spec.java</include>
                     </includes>
+                    <!-- Controller specs assert Hibernate Validator's default constraint
+                         messages, which are resolved against the JVM locale — pin English
+                         so the suite passes on non-English machines too. -->
+                    <argLine>@{argLine} -Duser.language=en -Duser.country=US</argLine>
```
UnLow1, line 140: The `@{argLine}` reference (late replacement, `@` not `$`) is load-bearing: JaCoCo's `prepare-agent` goal publishes the agent's `-javaagent` flag through the `argLine` property, and overriding `argLine` here without re-including it would silently detach the coverage agent — coverage would drop to 0% and the Codecov upload would report an empty run. Verified locally that `target/jacoco.exec` is still produced with this configuration.

The locale pin itself exists because the new specs assert Hibernate Validator's default messages ("must not be null" etc.), which resolve against the JVM locale — HV ships a Polish bundle, so an unpinned run on a `pl_PL` machine would fail these tests.

---

src/main/java/com/jamex/refereestaffer/controller/RestExceptionHandler.java
```diff
@@ -61,4 +67,44 @@ public ProblemDetail handleDownload(DownloadFileException ex) {
...
+        return ProblemDetail.forStatusAndDetail(HttpStatus.BAD_REQUEST, detail);
+    }
+
+    /**
+     * Validation failure on container elements — {@code List<@Valid Dto>} bodies (bulk match
+     * update, configuration update). Spring routes those through built-in method validation,
+     * which throws this instead of {@link MethodArgumentNotValidException}.
+     */
+    @ExceptionHandler(HandlerMethodValidationException.class)
+    public ProblemDetail handleHandlerMethodValidation(HandlerMethodValidationException ex) {
```
UnLow1, line 93: Two handlers instead of one is deliberate: a single `@Valid DtO` body fails as `MethodArgumentNotValidException` (thrown by the argument resolver's DataBinder), while `List<@Valid Dto>` bodies go through Spring 6.1+ built-in *method* validation, which throws this exception instead. They expose errors through different APIs (`BindingResult` vs `ParameterValidationResults`), hence the near-duplicate bodies sharing only `describe()`. If Spring ever unifies these, this collapses to one handler.

---

src/main/java/com/jamex/refereestaffer/model/dto/RefereeDto.java
```diff
@@ -1,20 +1,22 @@
...
 import jakarta.validation.constraints.NotNull;
 
 public class RefereeDto {
 
-    @NotNull
+    @NotNull(groups = OnUpdate.class)
     private final Long id;
 
-    @NotNull
+    @NotBlank
```
UnLow1, line 13: Behaviour change beyond the ticket's literal scope, flagging for visibility: `@NotNull` → `@NotBlank` means whitespace-only strings are now rejected too, not just missing ones (same for `lastName`, `email`, and `TeamDto.name`/`city`). The form drawers already prevent empty submissions, so no UI flow is affected — this only tightens direct API calls. Covered by the "should reject team creation when name or city is blank" spec, which posts `"name": "  "`.

---

UnLow1 (commented):
Self-review notes on the more discussion-worthy spots of this change — see inline comments (bulk-endpoint validation gap, entity-level validation side effect on Config, the JaCoCo argLine interplay, the dual exception handlers, and the @NotBlank tightening).

---

src/main/java/com/jamex/refereestaffer/controller/MatchController.java
UnLow1: Addressed in 9d9224c — `updateMatches` now checks id presence explicitly (`requireIds`) and a missing id 400s with the same `[index].id: must not be null` detail format, via a new `RequestValidationException` mapped in `RestExceptionHandler`. Covered by the "should reject bulk match update when an element has no id" spec.

---

PR #71, opened 2026-07-07
RS-69: Cover redesign-added backend logic with Spock specs

UnLow1:
## Ticket

[RS-69 — Testy logiki backendu dodanej przy redesignie](https://referee-staffer.youtrack.cloud/issue/RS-69)

The `design-refresh` branch added domain logic to the backend without any spec coverage — the biggest test gap on the backend, since this logic runs on every visit to the Staffer / match-detail / referee screens. Three areas needed tests: `MatchService.computeDifficultyBreakdown` (per-component hardness breakdown for the Staffer drawer and Match detail panels), `RefereeService.enrichWithStats` (the `P = α·avg + β·experience` potential formula with its `DEFAULT_GRADE` fallback), and the `homeWins`/`awayWins` counters in `RefereeService.calculateStats`.

## Plan

1. Extend the existing `MatchServiceSpec` with features for `computeDifficultyBreakdown`:
   - `MatchNotFoundException` for a missing match id,
   - a data-driven full-breakdown feature asserting that the parts sum to `total`, the flags mirror the parts, and `top`/`bottom` are mutually exclusive,
   - the unranked-team guard (`place == 0`) — stubbed config map deliberately omits the edge keys so a broken guard would NPE,
   - a feature proving standings (points and places) are recomputed from finished matches *inside* the call before scoring.
2. Extend the existing `RefereeServiceSpec`:
   - `enrichWithStats` computes potential from the average grade and experience using the two config multipliers,
   - `enrichWithStats` falls back to `DEFAULT_GRADE` for a referee with no graded matches,
   - `homeWins`/`awayWins`: draws and unfinished matches (including a single-`null` score) count for neither side; zero counters for a referee with no matches.
3. No production code changes; follow the existing spec conventions (Spock `Mock()`s, interaction stubbing in `then:` blocks, `where:` tables).
4. Verify with `mvn -DskipFrontend=true test`.

## Changes

- `MatchServiceSpec.groovy` — 4 new features (10 test executions) covering `computeDifficultyBreakdown`. The full-breakdown feature reuses the same `where:` table shape as the existing `getMatchesToAssignInQueue` feature so the two entry points are exercised against the same scenarios. The standings-refresh feature feeds a finished match through the mocked repository and asserts both the mutated `Team` points/places and the resulting `pointsDiff`/`base` part, proving the breakdown is computed against fresh standings rather than stale `@Transient` values.
- `RefereeServiceSpec.groovy` — 4 new features covering the win counters and `enrichWithStats`. Config multipliers are stubbed via `findByName` (matching how `enrichWithStats` reads them, as opposed to the `findAllAsMap` path used by match scoring).

Design decisions: asserted `top`/`bottom` mutual exclusivity explicitly (documented invariant of `DifficultyBreakdownDto`), and kept the NPE-trap technique from the existing unranked-guard spec for the breakdown variant so both entry points guard consistently.

## Testing

- `mvn -DskipFrontend=true test` — 130 tests, 0 failures, `BUILD SUCCESS` (JDK 25, Spock). New totals: `MatchServiceSpec` 27 executions, `RefereeServiceSpec` 8.

https://claude.ai/code/session_01TzqQA4DupN2cfedZKcDnNj

---
_Generated by [Claude Code](https://claude.ai/code/session_01TzqQA4DupN2cfedZKcDnNj)_

### DIFF
--- a/src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
+++ b/src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
@@ -163,4 +163,143 @@ class MatchServiceSpec extends Specification {
         1             | "city1"      | 2         | true    | true       | false
         3             | "city1"      | 2         | true    | false      | true
     }
+
+    def "should throw MatchNotFoundException when computing breakdown for missing match"() {
+        given:
+        def matchId = 44l
+
+        when:
+        matchService.computeDifficultyBreakdown(matchId)
+
+        then:
+        1 * matchRepository.findById(matchId) >> Optional.empty()
+        def exception = thrown(MatchNotFoundException)
+        exception.message == String.format(MatchNotFoundException.NOT_FOUND, matchId)
+    }
+
+    def "should compute difficulty breakdown with parts summing to total"() {
+        given:
+        def matchId = 7l
+        def homeTeam = [points: 10, place: 2, city: "city1"] as Team
+        def awayTeam = [points: 30, place: awayTeamPlace, city: awayTeamCity] as Team
+        def match = Match.builder().id(matchId).home(homeTeam).away(awayTeam).build()
+        def matchHardnessLvlMultiplier = 2.5d
+        def matchHardnessIncrementer = 100.0d
+        def matchHardnessDerbyIncrementer = 15.0d
+        def matchHardnessTopIncrementer = 11.0d
+        def matchHardnessBottomIncrementer = 7.0d
+
+        when:
+        def result = matchService.computeDifficultyBreakdown(matchId)
+
+        then:
+        1 * matchRepository.findById(matchId) >> Optional.of(match)
+        1 * matchRepository.findAllByHomeScoreNotNullAndAwayScoreNotNull() >> []
+        1 * configurationRepository.findAllAsMap() >> [
+                (ConfigName.DIFFICULTY_LEVEL_MULTIPLIER)                : matchHardnessLvlMultiplier,
+                (ConfigName.DIFFICULTY_LEVEL_INCREMENTER)               : matchHardnessIncrementer,
+                (ConfigName.NUMBER_OF_EDGE_TEAMS)                       : edgeTeams as Double,
+                (ConfigName.DIFFICULTY_LEVEL_SAME_CITY_INCREMENTER)     : matchHardnessDerbyIncrementer,
+                (ConfigName.DIFFICULTY_LEVEL_MATCH_ON_TOP_INCREMENTER)  : matchHardnessTopIncrementer,
+                (ConfigName.DIFFICULTY_LEVEL_MATCH_ON_BOTTOM_INCREMENTER): matchHardnessBottomIncrementer
+        ]
+        1 * teamRepository.count() >> 3
+
+        result.matchId() == matchId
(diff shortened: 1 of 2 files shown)


src/test/groovy/com/jamex/refereestaffer/service/RefereeServiceSpec.groovy
```diff
@@ -109,6 +111,101 @@ class RefereeServiceSpec extends Specification {
...
+        def referee = Referee.builder()
+                .firstName("John")
+                .lastName("Smith")
+                .experience(10)
+                .build()
+        def team1 = Team.builder().name("team1").build()
+        def team2 = Team.builder().name("team2").build()
+        def grade1 = 8.0
+        def grade2 = 9.0
+        def matches = createMatches(referee, team1, team2, team2, grade1, grade2)
```
UnLow1, line 172: Deliberate reuse of `team2` for the helper's `team3` parameter — this feature only asserts `averageGrade` and `potential`, not the `teamsRefereed` map, so two distinct teams are enough. If a third team ever matters here, switch to a dedicated `team3` instead of widening the helper.

---

src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
```diff
@@ -163,4 +163,143 @@ class MatchServiceSpec extends Specification {
...
+        result.total() == result.parts().base() + result.parts().sameCity() + result.parts().top() + result.parts().bottom()
+
+        and: "top and bottom are mutually exclusive"
+        !(result.parts().top() > 0 && result.parts().bottom() > 0)
+
+        and: "flags mirror the parts"
+        result.flags().sameCity() == isDerby
+        result.flags().isTop() == isTopMatch
+        result.flags().isBot() == isBottomMatch
+        result.flags().pointsDiff() == 20
```
UnLow1, line 224: The literal `20` is `|30 - 10|` from the fixed team points above (same convention as the existing `getMatchesToAssignInQueue` feature, which computes it inline). Kept as a literal so the flag assertion stays independent of the arithmetic it is checking.

---

src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
```diff
@@ -163,4 +163,143 @@ class MatchServiceSpec extends Specification {
...
+        !result.flags().isBot()
+
+        where:
+        homePoints | homePlace | awayPoints | awayPlace
+        0          | 0         | 0          | 0
+        0          | 0         | 30         | 1
+        30         | 1         | 0          | 0
+    }
+
+    def "should refresh standings from finished matches before computing breakdown"() {
```
UnLow1, line 271: This is the one feature where teams start with no `points`/`place` set: it pins down that `computeDifficultyBreakdown` recomputes standings from the finished matches *inside* the call (the ticket's "guard place == 0 / fresh standings" concern). It works because `calculatePointsForTeams` mutates the same `Team` instances referenced by both the finished match and the match under scoring — the assertions on `team1.place`/`team2.place` would fail if the service ever switched to detached copies.

---

UnLow1 (commented):
Self-review notes — context for the discussion-worthy spots in the diff.

---

PR #72, opened 2026-07-07
RS-79: Migrate to Angular 22, TypeScript 6, and the esbuild application builder

UnLow1:
# Ticket

[RS-79 — Migracja Angular 22 + TypeScript 6 + usunięcie bloku overrides](https://referee-staffer.youtrack.cloud/issue/RS-79)

Angular 22.0.5 is out and the repo policy is to take majors via `ng update` schematics rather than merging Dependabot bumps (PRs #67 and #40 were deliberately left hanging as reminders). The migration also unblocks TypeScript 6 (previously pinned out by `@angular-devkit/build-angular@21`'s `>=5.9 <6.0` peer range) and was the designated moment to retire the `package.json` `overrides` block that patched vulnerable transitive deps inside Angular 21's webpack dev-toolchain.

# Plan

1. `npm ci`, then `ng update @angular/core@22 @angular/cli@22 angular-eslint` (angular-eslint had to ride along — its v21 peer range blocks CLI 22).
2. Bump TypeScript to 6.x once the new toolchain's peer range allows it, plus `@types/node` major.
3. Remove the `overrides` block, reinstall, and verify with `npm audit` whether the patched advisories are actually gone from the new toolchain.
4. Re-verify the `ng serve` proxy (`proxy.conf.json`) — the nested `http-proxy-middleware` override existed precisely because a global one broke dev proxying (commit `36ac6f8`), so this is the known regression spot.
5. Review Node versions in `frontend-maven-plugin` and CI against Angular 22's requirement.
6. Gate everything on `npm run lint`, `npm run build`, `npm test` (headless), update `CLAUDE.md`, open PR.

The plan changed in one significant way during step 3: after removing the overrides, `npm audit` still reported all 8 advisories, because even the latest `@angular-devkit/build-angular@22.0.5` ships the vulnerable webpack dev-toolchain (webpack-dev-server → sockjs → uuid, http-proxy-middleware 3.0.x). The overrides weren't obsolete — the webpack builder itself was. So the fix became: migrate to the new build system via the opt-in `use-application-builder` migration, which drops the webpack toolchain entirely.

# Changes

**Angular 22 via `ng update`** — standard migrations applied:
- Every component got an explicit `changeDetection: ChangeDetectionStrategy.Eager` — this is the migration preserving pre-v22 behavior, since v22 flips the default to `OnPush`. A deliberate app-wide OnPush migration is a separate piece of work (follow-up candidate); until then the new `@angular-eslint/prefer-on-push-component-change-detection` rule (on by default in angular-eslint 22, 30 errors) is disabled in `eslint.config.js` with a comment.
- `provideHttpClient(withXhr(), ...)` in `main.ts` — preserves the XHR backend (v22 defaults to fetch).
- The migration also dropped `strictTemplates: false` + extended-diagnostic suppressions into the tsconfigs as behavior-preserving opt-outs. Instead of keeping them, `strictTemplates` (the v22 default) is now **enabled** — it surfaced exactly one real issue: `dashboard.component.html` guarded `hardnessLvl` (typed `number | undefined`) with `!== null` only. Fixed with `!= null`, and the template `eqeqeq` lint rule got its standard `allowNullOrUndefined` option.

**New build system (`@angular/build:application`, esbuild + Vite):**
- `use-application-builder` migration replaced all `@angular-devkit/build-angular` builders. Two manual follow-ups: `outputPath` is set to `{ base: "../resources/static", browser: "" }` so the bundle keeps landing flat where Spring Boot serves it (the new builder would otherwise nest it in `browser/`), and `karma.conf.js` lost the removed `build-angular` plugin/framework entries.
- `src/test.ts` deleted — the `@angular/build:karma` builder initializes the TestBed environment itself; `zone.js/testing` moved into the test target's `polyfills` array. The `teardown: {destroyAfterEach: false}` customization it carried dies with it — the full post-rebase suite passes under the default.

**TypeScript 6.0** — `baseUrl` removed from `tsconfig.json` (deprecated in TS 6, errors under the new toolchain; no `paths` depended on it). The migration itself dropped `downlevelIteration` and added `esModuleInterop`.

**Security state (`npm audit`): 8 advisories (2 high) → 1 low.** Dropping webpack retired the http-proxy-middleware and sockjs→uuid overrides. One minimal override remains — `@babel/core: ^7.29.7` — because `@angular/build@22.0.5` pins exact `@babel/core@7.29.0` (GHSA-4x5r-pxfx-6jf8); it should be dropped once Angular bumps upstream. The single remaining low advisory is Vite's nested `esbuild` (GHSA-g7r4-m6w7-qqqr): dev-server-only, Windows-only, and not fixable without forcing esbuild outside Vite's `^0.27.0` range — accepted and documented in `CLAUDE.md`.

**Not changed:** Node 24.15.0 in `frontend-maven-plugin` and `frontend.yml` — it is exactly Angular 22's minimum for the Node 24 line, so both stay as-is. `@types/node` bumped to ^26.

# Testing

- Rebased onto master after RS-80 (service specs) and RS-81 (component specs) merged — clean rebase, no conflicts, no new components to migrate (both PRs added specs only, plus one interceptor tweak).
- `npm run lint` — clean (0 errors).
- `npm run build` — clean, output verified to land flat in `src/main/resources/static` (index.html, main.js at the top level, no `browser/` nesting).
- `npm test -- --watch=false --browsers=ChromeHeadlessNoSandbox --no-progress` (the exact CI invocation) — **196/196 SUCCESS** on headless Chromium, including all RS-80/RS-81 specs running under the new `@angular/build:karma` builder and the default `destroyAfterEach: true` teardown.
- `ng serve` proxy regression check: started a stub JSON server on :8080, ran `npm start`, and confirmed `http://localhost:4200/api/team/all` proxies through the new Vite dev server to the backend.
- `npm audit` — 1 low (down from 3 low / 3 moderate / 2 high), see above.
- Backend untouched, so no Spock run; the Maven `npm install` + `npm run build` steps used by `frontend-maven-plugin` are the same commands verified above.

After merge, Dependabot PRs #67 (Angular 22) and #40 (TS 6) can be closed — both are superseded by this migration.

https://claude.ai/code/session_01KYF1qh3PcV53KDpxjj7Qg3


src/main/webapp/angular.json
```diff
@@ -11,12 +11,16 @@
       "schematics": {},
       "architect": {
         "build": {
-          "builder": "@angular-devkit/build-angular:browser",
+          "builder": "@angular/build:application",
           "options": {
-            "outputPath": "../resources/static",
+            "outputPath": {
+              "base": "../resources/static",
+              "browser": ""
```
UnLow1, line 18: Load-bearing: the `application` builder nests browser output in a `browser/` subfolder by default, which would silently move the bundle to `src/main/resources/static/browser/` and Spring Boot would serve a blank app. `browser: ""` flattens it back. If SSR is ever added this needs revisiting (server output would collide with the flattened browser output).

---

src/main/webapp/eslint.config.js
```diff
@@ -31,6 +31,9 @@ module.exports = defineConfig([
           style: "kebab-case",
         },
       ],
+      // The v22 update migration pinned every component to ChangeDetectionStrategy.Eager
+      // (the pre-v22 default). Re-enable this rule when the app migrates to OnPush.
+      "@angular-eslint/prefer-on-push-component-change-detection": "off",
```
UnLow1, line 36: Judgement call: the alternative was to accept the v22 `OnPush` default across all 29 components instead of the migration's explicit `Eager` pins. I kept the behavior-preserving route because the component layer has essentially no test coverage yet (RS-81) — flipping change detection app-wide with no safety net felt like a separate, deliberate piece of work. Once components migrate to OnPush (most state is already signals, so it's plausible), delete this rule override and drop the `Eager` pins in the same PR.

---

src/main/webapp/package.json
```diff
@@ -10,50 +10,42 @@
...
-    "@angular-devkit/build-angular": {
-      "http-proxy-middleware": "^3.0.7",
-      "webpack-dev-server": {
-        "http-proxy-middleware": "^2.0.9"
-      }
-    },
-    "sockjs": {
-      "uuid": "^11.1.1"
-    }
+    "@babel/core": "^7.29.7"
```
UnLow1, line 13: The ticket asked for the overrides block to be removed entirely, but this one entry has to stay for now: `@angular/build@22.0.5` pins exact `@babel/core@7.29.0`, which is inside the GHSA-4x5r-pxfx-6jf8 range (fixed in 7.29.7). Semver-compatible, so the override is safe. Re-run `npm audit` on the next Angular patch bump — the moment upstream moves past 7.29.0, this block can go.

---

src/main/webapp/src/app/component/dashboard/dashboard.component.html
```diff
@@ -54,7 +54,7 @@ <h1>Overview</h1>
             <td class="vs-cell muted mono">vs</td>
             <td><app-team-pill [team]="getTeam(match.awayTeamId)" side="away"></app-team-pill></td>
             <td class="num right col-difficulty">
-              @if (match.hardnessLvl !== null) {
+              @if (match.hardnessLvl != null) {
```
UnLow1, line 57: This is the single violation `strictTemplates` (now on — the v22 default) found in the whole codebase: `hardnessLvl` is typed `number | undefined` in the model, but the guard only checked `!== null`, so `undefined` could reach `app-meter`'s non-optional `value` input. The loose `!= null` covers both and narrows the type; the template `eqeqeq` rule got the standard `allowNullOrUndefined` option to permit exactly this idiom.

---

src/main/webapp/src/test.ts
```diff
@@ -1,13 +0,0 @@
-// This file is required by karma.conf.js and loads recursively all the .spec and framework files
```
UnLow1, line 1: One subtle behavior change hides in this deletion: the old manual `initTestEnvironment` set `teardown: { destroyAfterEach: false }`; the new builder's automatic init uses the Angular default (`true`, since v13). With the suite at one trivial pipe spec this changes nothing today, but when component specs get written for RS-80/RS-81, fixtures will be destroyed after each test — which is the behavior you want anyway.

---

UnLow1 (commented):
Self-review notes on the discussable decisions in this migration — details inline.

---

src/main/webapp/src/test.ts
UnLow1: Resolved by the rebase onto master: RS-80/RS-81 landed in the meantime with 195 new specs (services via HttpTestingController + the component layer), and the full suite — 196 tests — passes under the new automatic TestBed init with the default `destroyAfterEach: true`. The concern is now verified rather than hypothetical.

---

PR #73, opened 2026-07-08
RS-80: Add HttpTestingController specs for the eight frontend services

UnLow1:
## Ticket

[RS-80 — Testy frontendu: 8 serwisów z HttpTestingController](https://referee-staffer.youtrack.cloud/issue/RS-80)

After the 2026-06 redesign the frontend was left with exactly one trivial spec (`exclude-value.pipe.spec.ts` asserting only `toBeTruthy`). The whole HTTP service layer — 8 services — was untested. This PR rebuilds the first test layer: a spec per service verifying its HTTP contract, plus error handling in collaboration with the global `httpErrorInterceptor`. As a side task from the ticket, the `ExcludeValuePipe` spec is rewritten to assert the actual exclusion logic.

## Plan

1. Add a `*.spec.ts` next to each service in `src/app/service/` (Team, Referee, Match, Grade, Staffer, Vacation, Configuration, Importer) using the modern provider-based TestBed setup: `provideHttpClient(withInterceptors([httpErrorInterceptor]))` + `provideHttpClientTesting()` with `HttpTestingController`.
2. Per service, cover every public method: request URL, HTTP verb, request body (including the `{ids}` wrapper for `byIds`, `FormData` for the importer, `null` body for the staffer POST) and response deserialization.
3. Per service, cover error paths: a ProblemDetail error (404/409/400 with a `detail` field → toast shows the domain message) and a bare 500 (→ generic `Request failed (HTTP 500)` toast), asserting both the toast and that the error is rethrown to the subscriber — i.e. the documented interceptor contract.
4. Rewrite `exclude-value.pipe.spec.ts` with assertions of the filtering behaviour.
5. Verify with `npm ci`, `npm run lint`, headless `npm test`.

## Changes

- 8 new spec files under `src/main/webapp/src/app/service/` (55 new tests). Design decisions:
  - The specs register the **real** `httpErrorInterceptor` in the TestBed chain, mirroring `main.ts`, instead of testing services against a bare `HttpClient`. This makes the error specs meaningful: they assert the observable UI contract (error toast via `ToastService` signal + rethrow) rather than interceptor internals, and they'd catch a regression where the interceptor swallows errors.
  - No standalone interceptor spec — its behaviour (status 0 / ProblemDetail `detail` / fallback message) is exercised through the service specs, which is how it runs in production.
  - Happy-path tests assert `HttpTestingController` request objects (method, body, headers where the service sets them explicitly) and flush typed responses to verify deserialization, including the `blob` response type of `ImporterService.downloadExampleFile` and multipart `FormData` fields of `postFile`.
  - `StafferService`'s ProblemDetail case uses 409 (staffing conflict) and `ImporterService`'s uses 400 (import failure) to match the domain errors those endpoints actually produce; the CRUD services use 404.
- `exclude-value.pipe.spec.ts` rewritten: filtering by id, no-match and no-argument passthrough, string ids, strict-equality semantics (`1` vs `'1'`), and `null`/`undefined` input.
- No production code changes.

## Testing

- `npm run lint` — clean.
- `npm test -- --watch=false --browsers=ChromeHeadlessNoSandbox --no-progress` — **59/59 SUCCESS** (53 service tests + 6 pipe tests, executed locally with `CHROME_BIN` pointing at the container's Chromium; same flags as `frontend.yml` CI).

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_018D3ki5YXcjNh96qgozqEDH)_

### DIFF
--- a/src/main/webapp/src/app/pipe/exclude-value.pipe.spec.ts
+++ b/src/main/webapp/src/app/pipe/exclude-value.pipe.spec.ts
@@ -1,8 +1,40 @@
 import { ExcludeValuePipe } from './exclude-value.pipe';
 
 describe('ExcludeValuePipe', () => {
-  it('create an instance', () => {
-    const pipe = new ExcludeValuePipe();
-    expect(pipe).toBeTruthy();
+  const pipe = new ExcludeValuePipe();
+  const items = [
+    {id: 1, name: 'first'},
+    {id: 2, name: 'second'},
+    {id: 3, name: 'third'},
+  ];
+
+  it('filters out the item with the given id', () => {
+    expect(pipe.transform(items, 2)).toEqual([
+      {id: 1, name: 'first'},
+      {id: 3, name: 'third'},
+    ]);
+  });
+
+  it('returns all items when no id matches', () => {
+    expect(pipe.transform(items, 99)).toEqual(items);
+  });
+
+  it('returns all items when no id is given', () => {
+    expect(pipe.transform(items)).toEqual(items);
+  });
+
+  it('works with string ids', () => {
+    const stringItems = [{id: 'a'}, {id: 'b'}];
+    expect(pipe.transform(stringItems, 'a')).toEqual([{id: 'b'}]);
+  });
+
+  it('does not exclude on a loosely-equal id of a different type', () => {
+    // strict !== on purpose: numeric 1 must not match string '1'
+    expect(pipe.transform(items, '1')).toEqual(items);
+  });
+
+  it('returns undefined for null or undefined input', () => {
+    expect(pipe.transform(null, 1)).toBeUndefined();
+    expect(pipe.transform(undefined, 1)).toBeUndefined();
   });
 });
--- a/src/main/webapp/src/app/service/importer.service.spec.ts
+++ b/src/main/webapp/src/app/service/importer.service.spec.ts
@@ -1,0 +1,91 @@
+import {TestBed} from '@angular/core/testing';
+import {HttpErrorResponse, provideHttpClient, withInterceptors} from '@angular/common/http';
+import {HttpTestingController, provideHttpClientTesting} from '@angular/common/http/testing';
+
+import {ImporterService} from './importer.service';
+import {ToastService} from './toast.service';
+import {httpErrorInterceptor} from './http-error.interceptor';
+import {ImportResponse} from '../request/importResponse';
+
+describe('ImporterService', () => {
+  const importerUrl = '/api/importer';
+
+  let service: ImporterService;
+  let httpTesting: HttpTestingController;
+  let toastService: ToastService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        provideHttpClient(withInterceptors([httpErrorInterceptor])),
+        provideHttpClientTesting(),
+      ],
+    });
+    service = TestBed.inject(ImporterService);
+    httpTesting = TestBed.inject(HttpTestingController);
+    toastService = TestBed.inject(ToastService);
+  });
+
+  afterEach(() => {
+    httpTesting.verify();
+  });
+
+  it('postFile POSTs the file and queue count as multipart form data', () => {
+    const file = new File(['queue;home;away'], 'season.csv', {type: 'text/csv'});
+    const response: ImportResponse = {matches: 240, referees: 20, grades: 180, teams: 16};
+    let result: ImportResponse | undefined;
+    service.postFile(file, 30).subscribe(r => result = r);
+
+    const req = httpTesting.expectOne(importerUrl);
+    expect(req.request.method).toBe('POST');
+    const body = req.request.body as FormData;
+    expect(body instanceof FormData).toBeTrue();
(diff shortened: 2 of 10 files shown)


src/main/webapp/src/app/service/team.service.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        // Same interceptor chain as in main.ts, so error specs exercise the real
+        // toast-and-rethrow behaviour, not a bare HttpClient.
+        provideHttpClient(withInterceptors([httpErrorInterceptor])),
+        provideHttpClientTesting(),
+      ],
+    });
```
UnLow1, line 26: This setup pattern (real `httpErrorInterceptor` in the chain + `provideHttpClientTesting()`) is repeated verbatim in all eight service specs. I kept it inline rather than extracting a shared `configureServiceTestBed()` helper — at ~6 lines per spec the duplication felt cheaper than a test-utils indirection, and each spec stays runnable/readable in isolation. Happy to extract a helper if you prefer once component specs (RS-81) start needing the same setup.

---

src/main/webapp/src/app/service/importer.service.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+      next: () => fail('expected an error'),
+      error: (e: HttpErrorResponse) => error = e,
+    });
+
+    httpTesting.expectOne(`${importerUrl}/example`)
+      .flush(new Blob(), {status: 500, statusText: 'Internal Server Error'});
+
+    expect(error?.status).toBe(500);
+    expect(toastService.toast()).toEqual({message: 'Request failed (HTTP 500)', kind: 'error'});
+  });
```
UnLow1, line 109: Note on this one: because `downloadExampleFile` requests `responseType: 'blob'`, an error body arrives as a `Blob`, so `error.error?.detail` in the interceptor is always undefined and the toast falls back to the generic `Request failed (HTTP 500)` message — even if the backend actually sent a ProblemDetail JSON. The test locks in that current behaviour. If we ever want the real domain message on failed downloads, the interceptor would need to read and parse the Blob (async) — out of scope for this ticket.

---

src/main/webapp/src/app/service/staffer.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      error: (e: HttpErrorResponse) => error = e,
+    });
+
+    httpTesting.expectOne(`${stafferUrl}/5`)
+      .flush({detail: 'Not enough referees available for queue 5'}, {status: 409, statusText: 'Conflict'});
+
+    expect(error?.status).toBe(409);
+    expect(toastService.toast()).toEqual({message: 'Not enough referees available for queue 5', kind: 'error'});
+  });
+
```
UnLow1, line 70: Used 409 here (and 400 in the importer spec) instead of the 404 the CRUD specs use — matching the domain errors these endpoints actually produce (staffing conflict / import failure). From the interceptor's perspective the status code is irrelevant as long as `detail` is present, but the fixtures read closer to real backend behaviour this way.

---

src/main/webapp/src/app/pipe/exclude-value.pipe.spec.ts
```diff
@@ -1,8 +1,40 @@
...
+
+  it('works with string ids', () => {
+    const stringItems = [{id: 'a'}, {id: 'b'}];
+    expect(pipe.transform(stringItems, 'a')).toEqual([{id: 'b'}]);
+  });
+
+  it('does not exclude on a loosely-equal id of a different type', () => {
+    // strict !== on purpose: numeric 1 must not match string '1'
+    expect(pipe.transform(items, '1')).toEqual(items);
+  });
```
UnLow1, line 34: This deliberately locks in the strict `!==` semantics: a numeric `1` does not match the string `'1'`. Worth being aware of if a caller ever feeds the pipe an id straight from a route param (always a string) against numeric entity ids — the pipe would silently exclude nothing. Current callers pass typed numbers, so this is documentation, not a bug.

---

UnLow1 (commented):
Self-review notes: context on the shared TestBed pattern, the blob-error limitation of the interceptor, the non-404 status choices, and the strict-equality lock-in in the pipe spec.

---

src/main/webapp/src/app/service/importer.service.spec.ts
UnLow1: Addressed in 3d5f9f3 — the interceptor now reads JSON-typed Blob error bodies (`application/problem+json`) asynchronously and surfaces the real `detail` message; non-JSON blobs keep the generic fallback. Covered by a new async spec here and the existing generic-500 test.

---

src/main/webapp/src/app/pipe/exclude-value.pipe.spec.ts
UnLow1: Verified the only caller (`match-form` team selects): options bind via `[ngValue]="team.id"`, so the pipe always receives numeric ids — no live bug, keeping the strict semantics as documented.

---

PR #74, opened 2026-07-08
RS-81: Add component specs for the redesigned UI layer

UnLow1:
## Ticket

[RS-81 — Testy komponentów: najpierw staffer i match-form (hot-spoty), potem reszta](https://referee-staffer.youtrack.cloud/issue/RS-81)

The redesigned UI layer (2026-06) shipped with no component tests at all — the only existing frontend spec was a trivial pipe instantiation check. This PR covers the ticket's full list, in its risk order:

1. **`staffer.component`** — the core feature of the app (generate/lock/swap/save on signals), previously zero coverage.
2. **`match-form`** — the grade save/update/delete decision tree carried over verbatim from the legacy routed form, the most fragile data-flow in the UI.
3. **Entity form drawers** (referee / team / vacation).
4. **Overlay atoms** (`drawer`, `form-drawer`, `confirm-dialog`).
5. **Lists** (referee / match / team / vacation) incl. the legacy deep links.
6. **`standings`, `vacation-list` statuses, `importer`, `shell`.**

## Plan

1. Branch off `origin/master`, no production-code changes — tests only.
2. Hot-spots first: `staffer.component.spec.ts` (TestBed, services as Jasmine spies — `of(...)` for happy paths, `Subject` wherever emission *timing* is under test: loading lifecycle, the drawer's stale-response race guard) and `match-form.component.spec.ts` (the full four-way grade decision tree, each branch asserting the single right grade operation and that `saved` emits only after the grade round-trip).
3. Then the rest of the ticket list:
   - **Forms**: init in add vs edit mode (model is a *copy*; enriched/backend-owned fields ride along via the payload spread), invalid-form short-circuit, save-vs-update dispatch; for referee-form also DOM-level `fakeAsync` checks that errors gate on touched/dirty, the strict email pattern rejects TLD-less addresses, and the footer submit enables only when everything is valid; for vacation-form the `endBeforeStart` rule (missing dates, strict inequality, submit block).
   - **Overlays**: render-nothing-while-closed, X/backdrop/Esc all emit, Esc only while open; form-drawer's submit `form`-attribute wiring + validity gating, dirty indicator, and the discard guard (clean close vs guarded close, confirm closes, cancel keeps editing, and a single document-level Escape dismisses only the guard — the drawer's `tryClose()` must no-op while the guard is up).
   - **Lists**: load + join assertions (match-list joins only the ids actually referenced and skips empty lookups), ranking/filtering computeds, delete-through-confirm (guard names the entity; service called only on confirm), deep links `/addX` and `/addX/:id` opening the drawer, and URL normalization on close only when deep-linked.
   - **Standings**: W/D/L/GF/GA derived from finished matches only, accent/danger zones with the small-league suppression, signed goal-difference formatting, segment filters, latest played queue.
   - **Vacation statuses** (past/active/upcoming with inclusive boundary days), **importer** upload lifecycle (guards, uploading flag, error surface + retry reset), **shell** (Admin nav group gated on the UI setting, temporary toggle labels, explainer toggle placement, theme button).
4. Verify with `npm ci`, `npm run lint`, headless Karma run; self-review the diff.

## Changes

15 new spec files, 137 tests total (136 new + 1 pre-existing pipe spec). No production code touched.

Design decisions:

- **Spies over `HttpTestingController`** — the services are one-liner HTTP wrappers; mocking at the service boundary keeps the specs about component behaviour, not URL strings, and matches how the backend Spock controller specs isolate their layer.
- **`Subject`-driven assertions for async ordering** — loading flags, the staffer drawer race guard, and the saved-emits-after-grade contract are all about *when* things happen, so those specs hold the response open instead of using `of(...)`.
- **`componentRef.setInput` for input changes after the first change-detection pass** — direct instance mutation leaves the view unmarked and trips `NG0100` on the dev-mode `checkNoChanges` pass (all seven form-drawer specs failed that way on the first run).
- **Stubbed `UiSettingsService`** (writable signals + spies) wherever it's injected, so specs don't touch `localStorage`/`document` theme state.
- **Route stubs per test** — lists get a minimal `ActivatedRoute` snapshot (`url[0].path` + `paramMap`), which is all the deep-link logic reads, and a `Router` spy to assert URL normalization without a real router.
- **Real dates for vacation statuses** — `status()` compares against today, so the fixtures build ISO dates relative to `new Date()` instead of pinning the clock.

## Testing

- `npm run lint` — clean.
- `CHROME_BIN=… ng test --browsers=ChromeHeadlessNoSandbox --watch=false` — **137/137 SUCCESS**.
- Backend untouched; `frontend.yml` CI runs the same lint + headless test commands.

https://claude.ai/code/session_01MdSsSkQfeWTTTa4R141b17

### DIFF
--- a/src/main/webapp/src/app/component/match-form/match-form.component.spec.ts
+++ b/src/main/webapp/src/app/component/match-form/match-form.component.spec.ts
@@ -1,0 +1,247 @@
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {NgForm} from '@angular/forms';
+import {of, Subject} from 'rxjs';
+import {MatchFormComponent} from './match-form.component';
+import {MatchService} from '../../service/match.service';
+import {TeamService} from '../../service/team.service';
+import {RefereeService} from '../../service/referee.service';
+import {GradeService} from '../../service/grade.service';
+import {Match} from '../../model/match';
+import {Team} from '../../model/team';
+import {Referee} from '../../model/referee';
+import {Grade} from '../../model/grade';
+
+describe('MatchFormComponent', () => {
+  let matchService: jasmine.SpyObj<MatchService>;
+  let teamService: jasmine.SpyObj<TeamService>;
+  let refereeService: jasmine.SpyObj<RefereeService>;
+  let gradeService: jasmine.SpyObj<GradeService>;
+
+  const teams: Team[] = [
+    {id: 1, name: 'Alfa', city: 'Krakow', points: 40},
+    {id: 2, name: 'Beta', city: 'Gdansk', points: 30}
+  ];
+
+  const referees: Referee[] = [
+    {id: 100, firstName: 'Jan', lastName: 'Kowalski', email: 'jan@example.com', experience: 10}
+  ];
+
+  function makeMatch(overrides: Partial<Match> = {}): Match {
+    return {
+      id: 7,
+      queue: 3,
+      homeTeamId: 1,
+      awayTeamId: 2,
+      date: new Date('2026-03-01T12:00:00'),
+      refereeId: 100,
+      gradeId: undefined,
+      homeScore: 2,
+      awayScore: 1,
+      ...overrides
+    } as Match;
+  }
+
+  const validForm = {valid: true} as NgForm;
+  const invalidForm = {valid: false} as NgForm;
+
+  beforeEach(async () => {
+    matchService = jasmine.createSpyObj('MatchService', ['save', 'update']);
+    teamService = jasmine.createSpyObj('TeamService', ['findAll']);
+    refereeService = jasmine.createSpyObj('RefereeService', ['findAll']);
+    gradeService = jasmine.createSpyObj('GradeService', ['findById', 'save', 'update', 'delete']);
+
+    teamService.findAll.and.returnValue(of(teams));
+    refereeService.findAll.and.returnValue(of(referees));
+
+    await TestBed.configureTestingModule({
+      imports: [MatchFormComponent],
+      providers: [
+        {provide: MatchService, useValue: matchService},
+        {provide: TeamService, useValue: teamService},
+        {provide: RefereeService, useValue: refereeService},
+        {provide: GradeService, useValue: gradeService}
+      ]
+    }).compileComponents();
+  });
+
+  function createComponent(match: Match | null): ComponentFixture<MatchFormComponent> {
+    const fixture = TestBed.createComponent(MatchFormComponent);
+    fixture.componentInstance.match = match;
+    fixture.detectChanges();
+    return fixture;
+  }
+
+  describe('initialization', () => {
+    it('loads teams and referees for the selects in add mode', () => {
+      const component = createComponent(null).componentInstance;
+
+      expect(component.teams).toEqual(teams);
+      expect(component.referees).toEqual(referees);
+      expect(component.editMode).toBeFalse();
+      expect(component.model).toEqual({} as Match);
+      expect(gradeService.findById).not.toHaveBeenCalled();
+    });
+
+    it('copies the edited match into the form model so edits do not leak into the list', () => {
+      const match = makeMatch();
+      const component = createComponent(match).componentInstance;
+
+      expect(component.editMode).toBeTrue();
+      expect(component.model).toEqual(match);
+      expect(component.model).not.toBe(match);
+
+      component.model.queue = 99;
+      expect(match.queue).toBe(3);
+    });
+
+    it('fetches the existing grade when the match references one', () => {
+      const grade: Grade = {id: 5, value: 8.1};
+      gradeService.findById.and.returnValue(of(grade));
+
+      const component = createComponent(makeMatch({gradeId: 5})).componentInstance;
+
+      expect(gradeService.findById).toHaveBeenCalledWith(5);
+      expect(component.grade).toEqual(grade);
+    });
+
+    it('shows the mode in the drawer title', () => {
+      const addFixture = createComponent(null);
+      expect((addFixture.nativeElement as HTMLElement).querySelector('.drawer__title')?.textContent)
+        .toContain('Add match');
+      addFixture.destroy();
+
+      const editFixture = createComponent(makeMatch());
+      expect((editFixture.nativeElement as HTMLElement).querySelector('.drawer__title')?.textContent)
+        .toContain('Edit match');
+    });
+  });
+
+  describe('submit in add mode', () => {
+    it('does nothing while the form is invalid', () => {
+      const component = createComponent(null).componentInstance;
+
+      component.onSubmit(invalidForm);
+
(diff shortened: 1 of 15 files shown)


src/main/webapp/src/app/component/staffer/staffer.component.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+  ];
+
+  function makeMatch(id: number, overrides: Partial<Match> = {}): Match {
+    return {
+      id,
+      queue: 1,
+      homeTeamId: 1,
+      awayTeamId: 2,
+      date: new Date('2026-03-01T12:00:00'),
+      refereeId: undefined,
```
UnLow1, line 44: The `as Match` cast is needed because the model types `refereeId`/`gradeId`/scores as required `number`, while the backend (and the component's own `match.refereeId != null` checks) treat them as nullable. It compiles here only because `tsconfig.json` doesn't enable `strict`. If the model interfaces ever get cleaned up to `refereeId?: number` etc., this cast (and the one in the match-form spec) can go away — could be a small follow-up ticket alongside the `Team.place` modelling note in RS-27.

---

src/main/webapp/src/app/component/match-form/match-form.component.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+      expect(gradeService.delete).not.toHaveBeenCalled();
+      expect(emitted).toEqual([updated]);
+    });
+
+    it('deletes the grade when its value was cleared', () => {
+      createInEditMode(5, {id: 5, value: 7.5});
+      const gradeDelete = new Subject<void>();
+      gradeService.delete.and.returnValue(gradeDelete);
+
+      component.grade.value = undefined as unknown as number;
```
UnLow1, line 226: This pins the *current* behaviour of the grade branch, which is truthiness-based: `isNewGradeAdded()` checks `grade.value`, so a grade of exactly `0` is indistinguishable from "no grade entered" — editing an existing grade down to 0 would **delete** it rather than update it. Real observer grades are presumably never 0, so this is likely fine, but it's an inherited quirk of the legacy decision tree, not a deliberate contract. If it ever gets fixed (e.g. `grade.value != null`), this spec and the add-mode one are the two that will need updating.

---

src/main/webapp/src/app/component/staffer/staffer.component.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+
+      expect(component.drawerMatch()).toEqual(target);
+      expect(matchService.getDifficultyBreakdown).toHaveBeenCalledWith(11);
+      expect(component.drawerBreakdown()?.matchId).toBe(11);
+    });
+
+    it('discards a stale breakdown response after switching to another match', () => {
+      const first = new Subject<DifficultyBreakdown>();
+      const second = new Subject<DifficultyBreakdown>();
+      matchService.getDifficultyBreakdown.and.returnValues(first, second);
```
UnLow1, line 267: `returnValues(first, second)` deliberately replaces the default `callFake` from `beforeEach` for this one spec: two separate cold `Subject`s let the test open the drawer for match B *before* match A's response arrives, then flush A first. This is the only spec covering the race guard in `openDrawer` (`if (this.drawerMatchId() === match.id)`) — if that guard is ever refactored to `switchMap` on a signal-driven stream, this test is the one that proves the behaviour survived.

---

UnLow1 (commented):
Self-review notes on spots worth a reader's attention — no blocking issues.

---

src/main/webapp/src/app/component/match-list/match-list.component.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+    router = jasmine.createSpyObj('Router', ['navigate']);
+
+    matchService.findAll.and.returnValue(of(matches));
+    teamService.findByIds.and.returnValue(of(teams));
+    refereeService.findByIds.and.returnValue(of(referees));
+    gradeService.findByIds.and.returnValue(of(grades));
+    // The match form (rendered when the drawer opens) loads these on init.
+    teamService.findAll.and.returnValue(of(teams));
+    refereeService.findAll.and.returnValue(of(referees));
+    gradeService.findById.and.returnValue(of(grades[0]));
```
UnLow1, line 66: Non-obvious coupling: this stub isn't for the list itself. When a deep-link test opens the edit drawer, the fixture renders a real `MatchFormComponent`, whose `ngOnInit` fetches the grade because the fixture match carries `gradeId: 500`. Without this line the spy returns `undefined` and `.subscribe` throws inside change detection. The alternative would be overriding the list's imports to swap the form for a stub component — heavier, and rendering the real drawer is closer to what the deep link actually does in the app.

---

src/main/webapp/src/app/component/vacation-list/vacation-list.component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    const month = String(d.getMonth() + 1).padStart(2, '0');
+    const day = String(d.getDate()).padStart(2, '0');
+    return `${d.getFullYear()}-${month}-${day}`;
+  }
+
+  function makeVacation(id: number, refereeId: number, startDate: string, endDate: string): Vacation {
+    return {id, refereeId, startDate, endDate} as unknown as Vacation;
+  }
+
+  const past = makeVacation(31, 1, isoDay(-10), isoDay(-5));
```
UnLow1, line 33: These fixtures deliberately use the real clock (`isoDay` offsets from today) because `status()` internally compares against `todayIso()` — there's no injection seam for the current date. The trade-off: a test that starts before local midnight and asserts after it could flip `active`/`past` for the ±1-day fixtures. With a full suite run under 3 seconds that's a once-in-years flake, so it didn't seem worth introducing a clock abstraction into the component just for the spec — but if it ever fires in CI, wrapping `todayIso` behind an injectable is the fix.

---

src/main/webapp/src/app/component/referee-form/referee-form.component.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      const input = el.querySelector(selector) as HTMLInputElement;
+      input.value = value;
+      input.dispatchEvent(new Event('input'));
+    }
+
+    type('#firstName', 'Jan');
+    type('#lastName', 'Kowalski');
+    type('#experience', '5');
+    // Angular's email validator would accept this; the stricter pattern must not.
+    type('#email', 'jan@localhost');
```
UnLow1, line 135: `jan@localhost` is the value that distinguishes the component's stricter `pattern` from Angular's built-in `email` validator (which accepts TLD-less addresses). If someone ever removes the `[pattern]` binding from the template thinking `type="email"` + `email` validation covers it, this is the assertion that will catch the regression.

---

UnLow1 (commented):
Self-review notes for the second commit (remaining components) — no blocking issues.

---

PR #80, opened 2026-07-09
RS-7: Support split observer grades with mean-based effective value

UnLow1:
## Ticket

[RS-7 — Co z łamanym ocenami?](https://referee-staffer.youtrack.cloud/issue/RS-7)

In Polish football, an observer grade can be a *split grade* ("łamana ocena") — two components such as `7.9/8.3` instead of a single number. The system only stored one `value` per grade, so split grades could not be recorded at all. The ticket settles the design: store both components and use their arithmetic mean (`7.9/8.3 → 8.1`) wherever a single number is needed.

## Plan

1. **Backend model**: keep the existing `value` column as the first (and for plain grades, only) component and add a nullable `secondValue` column to `Grade`; expose `getEffectiveValue()` returning the mean when the second component is present. Keeping the existing column name avoids a data-losing rename under `ddl-auto: update`.
2. **API**: carry `secondValue` through `GradeDto` and `GradeConverter` (additive, backward-compatible JSON change).
3. **Statistics**: `RefereeService.countAverageGrade` averages effective values, so split grades feed referee potential correctly.
4. **Importer**: accept `7.9/8.3` in the CSV grade column; malformed grades (more than two components) fail the import with the existing 400 path.
5. **Frontend**: mirror the model (`secondValue?` + `effectiveGradeValue()` helper), add an optional second grade input to the match form, and switch every display/aggregation site (grade list, match list, referee profile) to the effective value, showing the `7.9/8.3` breakdown where relevant.
6. **Tests**: Spock for the entity, converter, referee average, importer, and controller JSON contract; Karma for the model helper and match-form behaviour.

## Changes

**Backend**

- `Grade` entity: new nullable `secondValue` column, `getEffectiveValue()` (mean for split grades, plain value otherwise, `null`-safe), extended builder/constructors/`toString`.
- `GradeDto` + `GradeConverter`: `secondValue` passes through both conversion directions; it stays optional (`@NotNull` only on `value`).
- `RefereeService.countAverageGrade`: uses `Grade::getEffectiveValue` instead of `Grade::getValue`.
- `ImporterService`: new `parseGrade` splits the grade column on `/`; one component → plain grade, two → split grade, more → `NumberFormatException`, which the existing catch wraps into `ImportException` (HTTP 400).
- Dev-profile seed: one demo grade is now a split grade so the feature is visible out of the box.

**Frontend**

- `model/grade.ts`: `secondValue?: number | null`, plus `effectiveGradeValue()` / `isSplitGrade()` helpers mirroring the backend semantics.
- Match form: the grade field is now a two-part input (`first / second`). The second input is disabled until the first component is filled and is cleared automatically when the first one is cleared, so a second-component-only grade can't be submitted. The save/update/delete decision tree in `onSubmit` is untouched.
- Grade list: meter, warn threshold, and the printed value use the effective value; split grades additionally show the `7.9/8.3` breakdown next to the mean.
- Match list + referee profile: meters, the match-history grade cell, and the client-side average fallback use the effective value.

**Design decisions**

- The mean is computed on read (`getEffectiveValue`) rather than stored, so there is a single source of truth and no denormalized column to keep in sync.
- `value` keeps its name and `NOT NULL` constraint; `secondValue` is additive — existing rows and API clients are unaffected, and H2 `ddl-auto: update` adds the column without touching data.

## Testing

- **Backend** (`mvn -DskipFrontend=true test`, JDK 25): 140 tests green. New/extended specs: `GradeSpec` (effective-value semantics incl. null value), `GradeConverterSpec` (split round-trip both directions), `RefereeServiceSpec` (average mixes split and plain grades via the mean), `ImporterServiceSpec` (imports `7.9/8.3`; `7.9/8.3/8.5` → `ImportException`), `GradeControllerSpec` (JSON contract for GET and POST with `secondValue`).
- **Frontend** (`npm ci`, `npm run lint`, `npm test` headless, `npm run build`): lint clean, 143 tests green, production build succeeds. New specs: `grade.spec.ts` (helper semantics incl. explicit `null`), match-form split-grade submit and second-component clearing.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01UadVCzrSLUPCJsQrkGPQrj)_

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/RefereeStafferApplication.java
+++ b/src/main/java/com/jamex/refereestaffer/RefereeStafferApplication.java
@@ -56,7 +56,7 @@ public class RefereeStafferApplication {
 
             var grade1 = new Grade(match1, 8.3);
             var grade2 = new Grade(match2, 8.2);
-            var grade3 = new Grade(match3, 7.9);
+            var grade3 = new Grade(match3, 7.9, 8.3);
             var grade4 = new Grade(match4, 8.5);
             var grade5 = new Grade(match5, 8.1);
             var grade6 = new Grade(match6, 8.4);
--- a/src/main/java/com/jamex/refereestaffer/model/entity/Grade.java
+++ b/src/main/java/com/jamex/refereestaffer/model/entity/Grade.java
@@ -19,6 +19,10 @@ public class Grade {
     @Column(nullable = false)
     private Double value;
 
+    // Second component of a "split" grade (e.g. 7.9/8.3). Null for a plain single grade.
+    @Column
+    private Double secondValue;
+
     @JsonIgnore
     @OneToOne(targetEntity = Match.class)
     private Match match;
@@ -26,9 +30,10 @@ public class Grade {
     public Grade() {
     }
 
-    public Grade(Long id, Double value, Match match) {
+    public Grade(Long id, Double value, Double secondValue, Match match) {
         this.id = id;
         this.value = value;
+        this.secondValue = secondValue;
         this.match = match;
     }
 
@@ -37,12 +42,33 @@ public class Grade {
         this.value = value;
     }
 
+    public Grade(Match match, double value, Double secondValue) {
+        this.match = match;
+        this.value = value;
+        this.secondValue = secondValue;
+    }
+
     public Long getId() {
         return id;
     }
 
     public Double getValue() {
         return value;
+    }
+
+    public Double getSecondValue() {
+        return secondValue;
+    }
+
+    /**
+     * The grade that counts towards referee statistics: the arithmetic mean of both
+     * components for a split grade (7.9/8.3 -> 8.1), or the single value otherwise.
+     */
+    public Double getEffectiveValue() {
+        if (value == null) {
+            return null;
+        }
+        return secondValue != null ? (value + secondValue) / 2 : value;
     }
 
     public Match getMatch() {
@@ -55,7 +81,7 @@ public class Grade {
 
     @Override
     public String toString() {
-        return "Grade(id=" + id + ", value=" + value + ")";
+        return "Grade(id=" + id + ", value=" + value + ", secondValue=" + secondValue + ")";
     }
 
     public static Builder builder() {
@@ -65,6 +91,7 @@ public class Grade {
     public static class Builder {
         private Long id;
         private Double value;
+        private Double secondValue;
         private Match match;
 
         public Builder id(Long id) {
@@ -77,13 +104,18 @@ public class Grade {
             return this;
         }
 
+        public Builder secondValue(Double secondValue) {
+            this.secondValue = secondValue;
+            return this;
+        }
+
         public Builder match(Match match) {
             this.match = match;
             return this;
         }
 
         public Grade build() {
-            return new Grade(id, value, match);
+            return new Grade(id, value, secondValue, match);
         }
     }
 }
--- a/src/main/java/com/jamex/refereestaffer/service/ImporterService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/ImporterService.java
@@ -111,8 +111,7 @@ public class ImporterService {
             matchRepository.save(match);
 
             if (line.length == 8 && queue <= numberOfQueuesToImport) {
-                var value = Double.parseDouble(line[7]);
-                var grade = new Grade(match, value);
+                var grade = parseGrade(match, line[7]);
                 gradeRepository.save(grade);
             }
         }
@@ -120,6 +119,18 @@ public class ImporterService {
         var matches = matchRepository.findAll();
         log.info(CREATED + grades.size() + " grades");
         log.info(CREATED + matches.size() + " matches");
+    }
+
+    // The grade column holds either a plain grade ("8.3") or a split grade ("7.9/8.3") —
+    // an observer grade broken into two components; referee stats use their mean.
+    private Grade parseGrade(Match match, String rawGrade) {
+        var parts = rawGrade.split("/");
+        if (parts.length > 2) {
+            throw new NumberFormatException("Invalid grade format: " + rawGrade);
+        }
+        var value = Double.parseDouble(parts[0]);
+        var secondValue = parts.length == 2 ? Double.parseDouble(parts[1]) : null;
+        return new Grade(match, value, secondValue);
     }
 
     private void createReferees(List<String> lines) {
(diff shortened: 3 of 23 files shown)


src/main/java/com/jamex/refereestaffer/model/entity/Grade.java
```diff
@@ -45,6 +56,21 @@ public Double getValue() {
...
+
+    /**
+     * The grade that counts towards referee statistics: the arithmetic mean of both
+     * components for a split grade (7.9/8.3 -> 8.1), or the single value otherwise.
+     */
+    public Double getEffectiveValue() {
+        if (value == null) {
+            return null;
+        }
+        return secondValue != null ? (value + secondValue) / 2 : value;
```
UnLow1, line 71: Deliberately not rounding here: `(7.9 + 8.3) / 2` is `8.100000000000001` in IEEE 754. Rounding is a presentation concern — every UI site already goes through `toFixed(1)` / `toFixed(2)`, and the referee-average pipeline should consume the unrounded mean to avoid compounding rounding error across matches. If a rounded value is ever needed on the API surface, it should be done in the DTO layer, not here.

---

src/main/java/com/jamex/refereestaffer/service/ImporterService.java
```diff
@@ -122,6 +121,18 @@ private void createMatchesAndGrades(List<String> lines, Short numberOfQueuesToIm
...
+    // an observer grade broken into two components; referee stats use their mean.
+    private Grade parseGrade(Match match, String rawGrade) {
+        var parts = rawGrade.split("/");
+        if (parts.length > 2) {
+            throw new NumberFormatException("Invalid grade format: " + rawGrade);
+        }
+        var value = Double.parseDouble(parts[0]);
+        var secondValue = parts.length == 2 ? Double.parseDouble(parts[1]) : null;
+        return new Grade(match, value, secondValue);
+    }
```
UnLow1, line 136: Edge-case behaviour of `String.split("/")` worth knowing here:

- `"8.3/"` → `["8.3"]` (trailing empty strings are dropped) → imported as a **plain** grade, silently. Arguably it should fail, but a trailing slash with nothing after it has an unambiguous reading, so I let it pass.
- `"/8.3"` → `["", "8.3"]` → `Double.parseDouble("")` throws `NumberFormatException` → wrapped into `ImportException` (400) by the existing catch in `importData`, same as `"7.9/8.3/8.5"`.

Both paths end in defined behaviour, so I didn't add extra guards beyond the `parts.length > 2` check.

---

src/main/webapp/src/app/component/match-form/match-form.component.html
```diff
@@ -80,8 +80,15 @@
...
-               [(ngModel)]="grade.value">
+        <div class="grade-split">
+          <input class="input num" type="number" id="grade" name="grade" min="0" step="0.1" placeholder="0.0"
+                 [(ngModel)]="grade.value" (ngModelChange)="onGradeValueChange($event)">
+          <span class="grade-split__sep mono">/</span>
+          <input class="input num" type="number" id="secondGrade" name="secondGrade" min="0" step="0.1" placeholder="—"
+                 title="Second component of a split grade (e.g. 7.9/8.3)"
+                 [(ngModel)]="grade.secondValue" [disabled]="gradeValueMissing">
+        </div>
+        <span class="hint">Split grade? Put the second component after the slash.</span>
```
UnLow1, line 91: Design choice: instead of cross-field validation (awkward in template-driven forms), the second input is simply **disabled** until the first component has a value, and clearing the first component auto-clears the second (`onGradeValueChange`). This makes the invalid state — a second component without a first — unrepresentable in the UI rather than merely rejected. The alternative (a custom validator directive) felt like overkill for one field pair; happy to revisit if more cross-field rules show up.

---

src/main/java/com/jamex/refereestaffer/RefereeStafferApplication.java
```diff
@@ -56,7 +56,7 @@ CommandLineRunner init(RefereeRepository refereeRepository, GradeRepository grad
 
             var grade1 = new Grade(match1, 8.3);
             var grade2 = new Grade(match2, 8.2);
-            var grade3 = new Grade(match3, 7.9);
+            var grade3 = new Grade(match3, 7.9, 8.3);
```
UnLow1, line 59: Strictly speaking this dev-seed tweak goes slightly beyond the ticket, but without it the feature is invisible when running with the `dev` profile — one split grade makes the breakdown rendering and the mean-based meter immediately verifiable by eye. Easy to drop if you prefer the seed untouched.

---

UnLow1 (commented):

---

src/main/java/com/jamex/refereestaffer/service/ImporterService.java
UnLow1: Tightened in 93e3d1e: `split("/", -1)` keeps trailing empty strings, so `"8.3/"` now fails on `Double.parseDouble("")` and is rejected with `ImportException` (400) like the other malformed forms, instead of silently importing as a plain grade. The spec is now parameterized over `"7.9/8.3/8.5"`, `"8.3/"`, and `"/8.3"` — all three rejected.

---

src/main/java/com/jamex/refereestaffer/model/entity/Grade.java
UnLow1: Keeping as is — rounding stays a presentation concern. The only consumers are `RefereeService.countAverageGrade` (wants the unrounded mean) and UI cells that already format with `toFixed`, so rounding in the entity would only introduce error with no benefit.

---

src/main/webapp/src/app/component/match-form/match-form.component.html
UnLow1: Keeping the disabled-input approach — it's covered by two match-form specs (second component dropped when the first is cleared, kept otherwise) and stays consistent with the repo's template-driven-forms convention. A validator directive would be the right move only if more cross-field rules accumulate.

---

src/main/java/com/jamex/refereestaffer/RefereeStafferApplication.java
UnLow1: Keeping the seed change — one split grade in the dev dataset is the cheapest way to make the feature visible end-to-end (grade list breakdown, meter, referee average) without importing a CSV. It only affects the `dev` profile's in-memory data.

---

PR #81, opened 2026-07-10
RS-14: Add client-side pagination to the match list

UnLow1:
## Ticket

[RS-14 — Page'owanie w liście meczów](https://referee-staffer.youtrack.cloud/issue/RS-14)

The match list rendered every filtered row at once (`GET /api/matches` has no paging). With a real season's worth of fixtures (240+ matches) the table gets heavy. The ticket calls for client-side pagination styled with the repo's own atom library (`component/common/`) — Material Paginator is no longer an option since Angular Material was removed in the redesign. Server-side paging (`Pageable`) is deliberately deferred until data actually grows.

## Plan

1. Build a reusable `app-paginator` atom in `component/common/paginator/` following the atom conventions (`@Input`/`@Output`, inline template, global styles in `styles.scss`, mono + `tabular-nums` for numbers), plus `chevLeft`/`chevRight` icons in `IconComponent`.
2. Integrate it into `MatchListComponent`: a `page` signal with `pageSize = 25`, a computed `pagedMatches` slice, page clamping when the filtered list shrinks, and a page reset on any filter change. Render the paginator in a `panel__foot` only when there is more than one page.
3. Cover it with tests: a spec for the new atom and new match-list spec cases for slicing, page reset, and clamping.

> Rebased onto current `master` (originally branched before RS-15/RS-79/RS-85 landed). RS-15 delivered the "All queues" seg option this PR initially added itself — that part was dropped in favour of RS-15's version (which also survives drawer-save reloads via the `undefined`/`null` sentinel split), and the page reset now additionally covers RS-15's result/referee status filters. The paginator also picked up the `ChangeDetectionStrategy.Eager` pin introduced by the Angular 22 migration (RS-79), and the footer sits below RS-85's `.tbl-scroll` wrapper.

## Changes

- **`PaginatorComponent`** (new atom): shows the visible range ("26–50 of 240"), current/total pages, and prev/next ghost buttons. Purely presentational — the owner slices its own list and reacts to `pageChange`. Guards against emitting pages outside `1..totalPages`.
- **`MatchListComponent`**: iterates `pagedMatches()` instead of `visibleMatches()`. The requested `page` signal is clamped by a `currentPage` computed (`min(page, totalPages)`), so deletes or filter changes can never strand the view on an empty page. Every filter mutation resets to page 1: `selectQueue()`, `setSearch()`, and the new `setResultFilter()`/`setRefereeFilter()` wrappers (the filter-bar segs previously called `resultFilter.set(...)` straight from the template) plus `clearFilters()`. The paginator footer renders only when the list exceeds one page, so short per-queue lists look exactly as before.
- **`IconComponent`**: `chevLeft`/`chevRight` lucide paths.
- **`styles.scss`**: `.pager` styles (mono, `tabular-nums`, ghost buttons), designed to sit in the existing `.panel__foot`.
- **`CLAUDE.md`**: `paginator` added to the atom-library list.

## Testing

- New `paginator.component.spec.ts`: range text (full, partial last page, zero items), disabled states at both ends, `pageChange` emissions, and the out-of-range guard.
- Extended `match-list.component.spec.ts`: page slicing over 60 matches, reset-to-first-page on queue/search/status-filter changes (including `clearFilters()`), and clamping when the filtered list shrinks.
- Full frontend pipeline re-run after the rebase on Node 24.15.0 (the version CI pins; Angular 22 rejects the sandbox's default Node 22.22.2): `npm ci`, `npm run lint` (clean), `npm test` headless (**257/257 green**), `npm run build` (esbuild builder, success).

🤖 Generated with Claude Code


src/main/webapp/src/app/component/match-list/match-list.component.ts
```diff
@@ -45,9 +46,14 @@ export class MatchListComponent implements OnInit {
   readonly refereesById = signal<Map<number, Referee>>(new Map());
   readonly gradesById = signal<Map<number, Grade>>(new Map());
 
+  /** `null` = all queues. */
   readonly selectedQueue = signal<number | null>(null);
   readonly searchTerm = signal('');
 
+  /** Requested page (1-based). May point past the end after the list shrinks — see `currentPage`. */
+  readonly page = signal(1);
+  readonly pageSize = 25;
```
UnLow1, line 64: Deliberately a fixed 25 with no page-size selector — the ticket asks for "client-side is enough for a start", and a 10/25/50 seg would be easy to bolt on later (the paginator already takes `pageSize` as an input). If reviewers prefer a selector now, it's a small follow-up.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
```diff
@@ -20,6 +20,11 @@ <h1>Matches</h1>
 <div class="panel">
   <div class="panel__head">
     <div class="seg">
+      @if (availableQueues().length > 0) {
+        <button type="button" [class.active]="selectedQueue() === null" (click)="selectQueue(null)">
+          All
+        </button>
+      }
```
UnLow1, line 50: Slight scope extension beyond the ticket: the component's filter always supported `selectedQueue === null` ("all queues"), but the UI never exposed it — so the whole-season list, the exact case pagination exists for, was unreachable. Adding "All" here makes the feature observable; team-name search also becomes season-wide when it's active. Default selection is unchanged (latest queue). Flagging in case this is considered RS-15 (filters) territory instead.

---

src/main/webapp/src/app/component/common/paginator/paginator.component.ts
```diff
@@ -0,0 +1,56 @@
...
+          <app-icon name="chevRight" [size]="12"></app-icon>
+        </button>
+      </div>
+    </div>
+  `
+})
+export class PaginatorComponent {
+  @Input({required: true}) total = 0;
+  /** 1-based. The owner is expected to keep it within 1..totalPages. */
+  @Input({required: true}) page = 1;
```
UnLow1, line 37: Design decision: the paginator is purely presentational and trusts the owner to pass an in-range `page` (match-list clamps via its `currentPage` computed). The alternative — the atom self-clamping and re-emitting — would give it two sources of truth and make the binding cycle harder to reason about. The `goTo` guard is the only defensive bit, so stray clicks can never emit an out-of-range page.

---

UnLow1 (commented):

---

src/main/webapp/src/app/component/match-list/match-list.component.ts
UnLow1: Resolution: keeping the fixed 25. RS-14 explicitly scopes this to "client-side is enough for a start", and per-queue lists (~8 rows) never hit the limit anyway — the selector would be UI with no current user. The atom already accepts `pageSize` as an input, so a 10/25/50 seg is a drop-in follow-up if the "All" view gets heavy use. No change.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
UnLow1: Resolution: keeping "All". Without it the paginator is effectively dead code — no reachable view exceeds ~8 rows — and RS-14's premise ("the table gets heavy with 240+ matches") only materializes on the whole-season list. RS-15 (real filters behind the placeholder Filter button) can build on top of this; it doesn't conflict. No change.

---

src/main/webapp/src/app/component/common/paginator/paginator.component.ts
UnLow1: Resolution: keeping the atom presentational. Single source of truth stays in the owning list (`page` signal + `currentPage` clamp), which matches how the other atoms work (`seg`, `meter` render what they're given). Both halves are covered by specs: the paginator's out-of-range guard and the match-list clamp after a shrink. No change.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
UnLow1: Superseded by the rebase: RS-15 (#82) landed on master with its own "All queues" option (including the `undefined`/`null` sentinel split so the choice survives drawer-save reloads), so this PR no longer adds it — the scope-extension concern is moot. This PR now only hooks the page reset into RS-15's status filters.

---

PR #82, opened 2026-07-11
RS-15: Wire up the match list Filter button with real status filters

UnLow1:
## Ticket

[RS-15 — Filtrowanie w liście meczów](https://referee-staffer.youtrack.cloud/issue/RS-15)

The match list already gained team-name search and a queue segment during the 2026-06 redesign, but the "Filter" button in the page head remained a disabled placeholder that did nothing. The ticket asked to either wire up real filters (match status: played / not played, with / without assignment, queue range) or remove the button. This PR wires it up.

## Plan

1. Keep everything client-side — the list already holds all matches in a signal, so filtering is a pure `computed()` extension; no backend changes needed.
2. Make the "Filter" button toggle an **inline filter bar** rather than a popover. The repo deliberately has only two overlay primitives (`app-drawer`, `app-confirm-dialog`), so no third overlay is introduced.
3. Put two segmented controls (`app-seg` atom) in the bar:
   - **Result**: All / Played / Not played — a match counts as played once both score halves are recorded (same rule `scoreOrPlaceholder` already uses).
   - **Referee**: All / Assigned / Unassigned — keyed on `refereeId` presence, matching the "unassigned" label already shown in the table.
4. Surface active-filter state on the button itself (count badge + accent border) so it stays visible when the bar is collapsed, and add a one-click "Clear" action.
5. Add an "All queues" option to the queue segment — the minimal take on the ticket's "queue range", and status filters are far more useful when they can span the whole season. `selectedQueue === null` already meant "no queue filter" in `visibleMatches`, it just had no UI.
6. Cover the new logic with component specs and run lint + the full Karma suite headless.

## Changes

- `match-list.component.ts`
  - New signals `filtersOpen`, `resultFilter`, `refereeFilter` plus `activeFilterCount` computed and `toggleFilters()` / `clearFilters()` methods.
  - `visibleMatches` now also filters by result and referee assignment; the comparisons are written as `(result === 'played') === isPlayed(m)` so one predicate covers both directions of each filter.
  - `selectQueue` accepts `null` for the "All queues" segment option.
  - `ResultFilter` / `RefereeFilter` string-literal union types keep the seg options and signals in sync.
- `match-list.component.html`
  - Filter button enabled: toggles the bar, shows a count badge, gets an accent border while any filter is active.
  - New `.filter-bar` block with two labelled `app-seg` groups and a ghost "Clear" button (rendered only when something is active).
  - "All queues" button prepended to the queue segment (only when there are queues at all).
  - Empty-state message distinguishes "no matches match the current filters" from the search / plain-queue cases.
- `match-list.component.scss` — styles for the filter bar, label, count badge and active button state, all on existing design tokens (border over shadow, single accent, `tabular-nums` on the numeric badge).
- `match-list.component.spec.ts` — six new specs (see Testing).

Design decisions: filters intentionally stay when the bar is collapsed (the badge tells you they're on); the search-term empty state keeps priority over the filter empty state since a typed search is the more specific signal.

## Testing

- `npm run lint` — clean.
- `npm test` (Karma, ChromeHeadlessNoSandbox) — **142/142 SUCCESS**, including new specs for: result + referee filtering across all queues, combining status filters with queue selection and team search, the half-recorded-score edge case counting as not played, `activeFilterCount` + `clearFilters()`, and the filter-bar toggle.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01MKaEzP6rbQJSwyABczR7Vt)_

### DIFF
--- a/src/main/webapp/src/app/component/match-list/match-list.component.html
+++ b/src/main/webapp/src/app/component/match-list/match-list.component.html
@@ -6,10 +6,12 @@
   <div class="page-head__actions">
     <input type="text" class="input" placeholder="Search team…"
            [value]="searchTerm()" (input)="setSearch($any($event.target).value)">
-    <!-- Filter button is a placeholder — advanced filters (referee assignment,
-         date range, etc.) are a future addition. -->
-    <button type="button" class="btn" disabled title="Filters coming soon">
+    <button type="button" class="btn" [class.btn--filter-active]="activeFilterCount() > 0"
+            (click)="toggleFilters()" title="Filter matches">
       <app-icon name="filter" [size]="12"></app-icon>Filter
+      @if (activeFilterCount() > 0) {
+        <span class="filter-count">{{ activeFilterCount() }}</span>
+      }
     </button>
     <button type="button" class="btn btn--primary" (click)="addMatch()">
       <app-icon name="plus" [size]="12"></app-icon>New match
@@ -17,9 +19,34 @@
   </div>
 </div>
 
+@if (filtersOpen()) {
+  <div class="filter-bar">
+    <div class="filter-bar__group">
+      <span class="filter-bar__label">Result</span>
+      <app-seg [options]="resultOptions" [value]="resultFilter()"
+               (valueChange)="resultFilter.set($event)"></app-seg>
+    </div>
+    <div class="filter-bar__group">
+      <span class="filter-bar__label">Referee</span>
+      <app-seg [options]="refereeOptions" [value]="refereeFilter()"
+               (valueChange)="refereeFilter.set($event)"></app-seg>
+    </div>
+    @if (activeFilterCount() > 0) {
+      <button type="button" class="btn btn--ghost btn--sm" (click)="clearFilters()">
+        <app-icon name="x" [size]="12"></app-icon>Clear
+      </button>
+    }
+  </div>
+}
+
 <div class="panel">
   <div class="panel__head">
     <div class="seg">
+      @if (availableQueues().length > 0) {
+        <button type="button" [class.active]="selectedQueue() === null" (click)="selectQueue(null)">
+          All queues
+        </button>
+      }
       @for (q of availableQueues(); track q) {
         <button type="button" [class.active]="q === selectedQueue()" (click)="selectQueue(q)">
           Queue {{ q }}
@@ -86,6 +113,8 @@
           <td colspan="6" class="empty-row">
             @if (searchTerm()) {
               <span class="muted">No matches match "{{ searchTerm() }}".</span>
+            } @else if (activeFilterCount() > 0) {
+              <span class="muted">No matches match the current filters.</span>
             } @else {
               <span class="muted">No matches in this queue.</span>
             }
--- a/src/main/webapp/src/app/component/match-list/match-list.component.html
+++ b/src/main/webapp/src/app/component/match-list/match-list.component.html
@@ -7,6 +7,7 @@
     <input type="text" class="input" placeholder="Search team…"
            [value]="searchTerm()" (input)="setSearch($any($event.target).value)">
     <button type="button" class="btn" [class.btn--filter-active]="activeFilterCount() > 0"
+            [attr.aria-expanded]="filtersOpen()"
             (click)="toggleFilters()" title="Filter matches">
       <app-icon name="filter" [size]="12"></app-icon>Filter
       @if (activeFilterCount() > 0) {
@@ -115,6 +116,8 @@
               <span class="muted">No matches match "{{ searchTerm() }}".</span>
             } @else if (activeFilterCount() > 0) {
               <span class="muted">No matches match the current filters.</span>
+            } @else if (selectedQueue() === null) {
+              <span class="muted">No matches yet.</span>
             } @else {
               <span class="muted">No matches in this queue.</span>
             }
--- a/src/main/webapp/src/app/component/match-list/match-list.component.scss
+++ b/src/main/webapp/src/app/component/match-list/match-list.component.scss
@@ -8,6 +8,46 @@
   width: 220px;
 }
 
+// Filter button gets the accent treatment while any filter is active, so the state
+// stays visible even when the bar itself is collapsed.
+.btn--filter-active {
+  border-color: var(--accent);
+  color: var(--accent-ink);
+}
+
+.filter-count {
+  min-width: 16px; height: 16px; padding: 0 4px;
+  display: inline-flex; align-items: center; justify-content: center;
+  border-radius: 8px;
+  background: var(--accent);
+  color: white;
+  font-size: 10px; font-weight: 600;
+  font-variant-numeric: tabular-nums;
+}
+
+.filter-bar {
+  display: flex;
+  align-items: center;
+  gap: 20px;
+  margin-bottom: 16px;
+  padding: 10px 14px;
+  border: 1px solid var(--line);
+  border-radius: var(--r);
+  background: var(--panel);
+}
+
+.filter-bar__group {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+}
+
+.filter-bar__label {
+  font-size: 12px;
+  font-weight: 500;
+  color: var(--ink-3);
+}
+
 .match-row {
   cursor: pointer;
 }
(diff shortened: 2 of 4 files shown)


src/main/webapp/src/app/component/match-list/match-list.component.ts
```diff
@@ -79,8 +103,12 @@ export class MatchListComponent implements OnInit {
   readonly visibleMatches = computed(() => {
     const queue = this.selectedQueue();
     const term = this.searchTerm().trim().toLowerCase();
+    const result = this.resultFilter();
+    const referee = this.refereeFilter();
     return this.matches()
       .filter(m => queue == null || m.queue === queue)
+      .filter(m => result === 'all' || (result === 'played') === isPlayed(m))
```
UnLow1: The boolean-equality form `(result === 'played') === isPlayed(m)` covers both filter directions with one predicate, but it is admittedly denser than an explicit `result === 'played' ? isPlayed(m) : !isPlayed(m)`. Happy to expand it if the terse version reads badly to anyone else — the spec `filters by result and referee assignment across all queues` pins the behavior either way.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
```diff
@@ -6,20 +6,47 @@ <h1>Matches</h1>
...
+    }
+  </div>
+}
+
 <div class="panel">
   <div class="panel__head">
     <div class="seg">
+      @if (availableQueues().length > 0) {
+        <button type="button" [class.active]="selectedQueue() === null" (click)="selectQueue(null)">
+          All queues
```
UnLow1, line 48: Deliberate: the list still *defaults* to the latest queue on load (unchanged behavior), "All queues" only becomes available as an explicit choice. With a full season loaded this renders every match in one table — acceptable today at this data size, but this is exactly where pagination (RS-14) should hook in later.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
```diff
@@ -86,6 +113,8 @@ <h1>Matches</h1>
           <td colspan="6" class="empty-row">
             @if (searchTerm()) {
               <span class="muted">No matches match "{{ searchTerm() }}".</span>
+            } @else if (activeFilterCount() > 0) {
+              <span class="muted">No matches match the current filters.</span>
```
UnLow1, line 118: Ordering choice: when both a search term and status filters are active and nothing matches, the search message wins — a typed term is the more specific signal and the one the user most likely just changed. The filter state is still visible via the badge on the Filter button.

---

src/main/webapp/src/app/component/match-list/match-list.component.scss
```diff
@@ -8,6 +8,46 @@
...
 }
 
+// Filter button gets the accent treatment while any filter is active, so the state
+// stays visible even when the bar itself is collapsed.
+.btn--filter-active {
+  border-color: var(--accent);
+  color: var(--accent-ink);
+}
+
+.filter-count {
```
UnLow1, line 18: `color: white` on the accent background mirrors what `.btn--primary` already does in `styles.scss`, so the badge stays legible in both themes (dark-theme accent is lighter, oklch 0.70). If the design system ever grows an `--on-accent` token, both spots should switch to it together.

---

UnLow1 (commented):

---

src/main/webapp/src/app/component/match-list/match-list.component.ts
UnLow1: Applied in f82d5cb — the inline boolean-equality expressions are gone, replaced by named `matchesResult` / `matchesReferee` helpers that spell out both directions of each filter explicitly. Lint clean, 142/142 specs still green.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
UnLow1: Keeping as is: the load-time default (latest queue) is unchanged, so nobody gets the full-season table without asking for it, and the data volume today (few hundred rows, client-side already holds them all) is well within what the table renders instantly. The real fix for scale is pagination, tracked separately as RS-14.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
UnLow1: Keeping the ordering: the search-term message names the exact string that produced zero rows, which is more actionable than a generic filter message, and the active-filter badge on the Filter button still signals the other constraint. No code change.

---

src/main/webapp/src/app/component/match-list/match-list.component.scss
UnLow1: Keeping as is — consistency with `.btn--primary` is the point; introducing an `--on-accent` token just for this badge would be scope creep for RS-15. If a second dark-on-accent case ever appears, that's the moment to add the token and migrate both.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
```diff
@@ -6,20 +6,47 @@ <h1>Matches</h1>
   <div class="page-head__actions">
     <input type="text" class="input" placeholder="Search team…"
            [value]="searchTerm()" (input)="setSearch($any($event.target).value)">
-    <!-- Filter button is a placeholder — advanced filters (referee assignment,
-         date range, etc.) are a future addition. -->
-    <button type="button" class="btn" disabled title="Filters coming soon">
+    <button type="button" class="btn" [class.btn--filter-active]="activeFilterCount() > 0"
+            (click)="toggleFilters()" title="Filter matches">
```
UnLow1, line 11: **A11y:** this button now toggles a disclosure (the filter bar) but exposes no state to assistive tech — a screen reader hears "Filter, button" with no way to tell whether the bar is open. Add `[attr.aria-expanded]="filtersOpen()"` so the toggle state is announced.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
```diff
@@ -86,6 +113,8 @@ <h1>Matches</h1>
           <td colspan="6" class="empty-row">
             @if (searchTerm()) {
               <span class="muted">No matches match "{{ searchTerm() }}".</span>
+            } @else if (activeFilterCount() > 0) {
+              <span class="muted">No matches match the current filters.</span>
             } @else {
               <span class="muted">No matches in this queue.</span>
```
UnLow1, line 122: **Nit:** with "All queues" selected (no search, no filters) an empty table still says "No matches in this queue." — misleading copy now that a no-queue view exists. Only reachable when the season has zero matches, but a `selectedQueue() === null` branch with "No matches yet." would keep it honest.

---

src/main/webapp/src/app/component/match-list/match-list.component.ts
```diff
@@ -132,14 +160,23 @@ export class MatchListComponent implements OnInit {
     });
   }
 
-  selectQueue(q: number): void {
+  selectQueue(q: number | null): void {
```
UnLow1, line 168: **Bug:** now that `null` is a legal user choice ("All queues"), it collides with the pre-existing default in `load()` (line 156): `if (queues.length > 0 && this.selectedQueue() == null) selectedQueue.set(queues[0])`. `onSaved()` re-runs `load()`, so a user who picked "All queues" and then saves a match from the drawer gets snapped back to the latest queue. The two states need to be distinguishable — e.g. `undefined` = not yet initialized, `null` = All queues (the `queue == null` guard in `visibleMatches` already covers both).

---

UnLow1 (commented):
Second review pass: one real bug (the "All queues" selection is lost on drawer save because `null` doubles as the uninitialized sentinel in `load()`), one a11y gap (`aria-expanded` missing on the disclosure toggle), one copy nit in the empty state.

---

src/main/webapp/src/app/component/match-list/match-list.component.ts
UnLow1: Fixed in 0679a4b: `selectedQueue` is now `signal<number | null | undefined>(undefined)` — `undefined` means "not initialized" and only that state triggers the latest-queue default in `load()`; `null` ("All queues") survives the reload after a drawer save. Covered by a new regression spec (`keeps an explicit "All queues" choice across the reload after a drawer save`), 143/143 green.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
UnLow1: Fixed in 0679a4b — `[attr.aria-expanded]="filtersOpen()"` added to the Filter button.

---

src/main/webapp/src/app/component/match-list/match-list.component.html
UnLow1: Fixed in 0679a4b — the empty state now has a `selectedQueue() === null` branch that says "No matches yet." instead of the per-queue copy.

---

PR #83, opened 2026-07-11
RS-27: Serve the computed league table from the standings endpoint

UnLow1:
# Ticket

[RS-27 — Standings liczone na backendzie (endpoint zamiast client-side)](https://referee-staffer.youtrack.cloud/issue/RS-27)

The Standings screen and the Team list both derived P/W/D/L/GF/GA (and "played") client-side from `GET /api/matches`, so each of those screens pulled the entire match list just to compute table stats. The ticket asks for a backend endpoint that returns the league table with computed statistics and `place`, and for the frontend to consume it instead of counting locally.

# Plan

1. **Backend — new response shape.** Add a `StandingsDto` record (following the `DifficultyBreakdownDto` record precedent): `afterQueue` (highest queue with a finished match, `null` before the season starts — feeds the "after queue N" subtitle) plus `rows`, where each row carries the `TeamDto` fields (`id`, `name`, `city`, `short`) and the computed season stats (`points`, `place`, `played`, `wins`, `draws`, `losses`, `goalsFor`, `goalsAgainst`).
2. **Backend — computation.** Rework `TeamService.getStandings()` to accumulate stats in one pass over finished matches, include every team (zeroed stats for teams without a finished match), sort by points desc → goal difference desc → goals scored desc → name, and assign `place = index + 1`. Reuse `MatchService`'s `POINTS_FOR_*` constants so the weights stay single-sourced. Drop the now-unneeded `MatchService` dependency and the `findAllByIdNotIn` two-query dance.
3. **Backend — controller.** `TeamController.getStandings()` returns the DTO directly (service already produces the read model; no converter step).
4. **Frontend — model & service.** New `Standing extends Team` + `Standings` wrapper interfaces; `TeamService.getStandings()` returns `Observable<Standings>`.
5. **Frontend — consumers.** `standings` and `team-list` stop fetching the match list; `dashboard`, `staffer` and `match-detail` switch to `.rows`, with `place` read from the row instead of being re-derived from the array index.
6. **Tests.** Rewrite `TeamServiceSpec` for the real computation (stats, tie-breakers, afterQueue, no-matches/no-teams edge cases), update the `TeamControllerSpec` standings feature for the new JSON shape, and update the Angular specs (standings, team-list, staffer) to the new mock shape.

Risks called out up front: the response shape of `/api/teams/standings` changes (all five consumers live in this repo and are updated in the same PR; there are no external consumers), and equal-points ordering changes from insertion order to explicit tie-breakers.

# Changes

**Backend**

- `StandingsDto` — new record with nested `Row`; `shortCode` serialises as `short` exactly like `TeamDto`, so team pills render from a row directly.
- `TeamService.getStandings()` now builds the full table itself: a private `TeamStats` accumulator (int fields internally, narrowed to `short` only at the DTO boundary, to avoid the compound-assignment narrowing trap already documented on `Team.addPoints`), a comparator implementing points → GD → GF → name, and `afterQueue` taken as the max queue of the finished matches. Every team from `teamRepository.findAll()` is ranked, so teams with no finished match now appear with zeroed stats at the bottom instead of relying on a separate `findAllByIdNotIn` query (removed from `TeamRepository` along with the old `// TODO probably this if could be removed`).
- `MatchService.calculatePointsForTeams` is untouched — the difficulty/staffing path still uses the transient `Team.points`/`place` fields. Deduplicating that (and removing the transient fields from the entity, the "modelowanie `Team.place`" note in the ticket) is deliberately left to RS-72-adjacent work; the two code paths share the points constants.

**Frontend**

- New `model/standing.ts` (`Standing extends Team`, `Standings` wrapper); `TeamService.getStandings(): Observable<Standings>`.
- `standings.component` — single HTTP call; `computeStats`/`latestPlayedQueue` deleted; rows, `place` and the "after queue N" subtitle come from the response; zone logic (top/bottom 3) stays client-side as pure presentation.
- `team-list.component` — single HTTP call; `countPlayed` and the `MatchService` dependency deleted; "Played" renders `row.played`.
- `staffer.component` / `match-detail.component` — consume `rows`; `place` lookups use the backend value instead of `index + 1`.
- `match-detail.component` — the "Team comparison" panel gains Wins / Draws / Losses / Goals for / Goals against rows, and the "requires a per-team season-stats endpoint" placeholder note is removed — this PR delivers that endpoint.

# Testing

- **Backend:** `mvn -DskipFrontend=true test` — all Spock specs green. `TeamServiceSpec` rewritten (full-table computation, tie-breaker chain, pre-season alphabetical table, empty league); `TeamControllerSpec` asserts the new JSON shape including the wire-format `short` field and that the converter is no longer touched on this endpoint.
- **Frontend:** `npm ci`, `npm run lint` (clean), `npm test` headless — 137/137 specs pass; `npm run build` also run to catch strict template errors. Standings/team-list specs now mock the wrapper response; the staffer spec's fixture carries backend `place`.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_013yBbY5mngAdXdxvrKWjrgV)_

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/model/dto/StandingsDto.java
+++ b/src/main/java/com/jamex/refereestaffer/model/dto/StandingsDto.java
@@ -1,0 +1,38 @@
+package com.jamex.refereestaffer.model.dto;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+/**
+ * The computed league table served by {@code GET /api/teams/standings}. Replaces the
+ * old {@code Collection<TeamDto>} response so the frontend no longer has to derive
+ * P/W/D/L/GF/GA client-side from the full match list.
+ *
+ * {@code afterQueue} is the highest queue with at least one finished match ("table
+ * after queue N" in the UI), or {@code null} before the season starts.
+ */
+public record StandingsDto(
+        Short afterQueue,
+        List<Row> rows
+) {
+    /**
+     * One table row. Carries the same team fields as {@link TeamDto} (so team pills
+     * render from a row directly) plus the season stats. `place` is 1-based table
+     * position — teams without a finished match sort to the bottom with zeroed stats.
+     */
+    public record Row(
+            Long id,
+            String name,
+            String city,
+            @JsonProperty("short") String shortCode,
+            short points,
+            short place,
+            short played,
+            short wins,
+            short draws,
+            short losses,
+            short goalsFor,
+            short goalsAgainst
+    ) {}
+}
--- a/src/main/java/com/jamex/refereestaffer/service/TeamService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/TeamService.java
@@ -1,51 +1,110 @@
 package com.jamex.refereestaffer.service;
 
+import com.jamex.refereestaffer.model.dto.StandingsDto;
+import com.jamex.refereestaffer.model.entity.Match;
 import com.jamex.refereestaffer.model.entity.Team;
 import com.jamex.refereestaffer.repository.MatchRepository;
 import com.jamex.refereestaffer.repository.TeamRepository;
 import org.springframework.stereotype.Service;
 
-import java.util.Collection;
 import java.util.Comparator;
-import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.IntStream;
 
 @Service
 public class TeamService {
 
+    /** Shared zeroed stats for teams without a finished match — never mutated. */
+    private static final TeamStats NO_MATCHES = new TeamStats();
+
     private final TeamRepository teamRepository;
     private final MatchRepository matchRepository;
-    private final MatchService matchService;
 
-    public TeamService(TeamRepository teamRepository, MatchRepository matchRepository, MatchService matchService) {
+    public TeamService(TeamRepository teamRepository, MatchRepository matchRepository) {
         this.teamRepository = teamRepository;
         this.matchRepository = matchRepository;
-        this.matchService = matchService;
     }
 
-    public Collection<Team> getStandings() {
-        var matches = matchRepository.findAllByHomeScoreNotNullAndAwayScoreNotNull();
-        matchService.calculatePointsForTeams(matches);
+    /**
+     * Builds the full league table from finished matches. Every team is included —
+     * teams without a finished match get zeroed stats and sort to the bottom. Points
+     * use the same weights as {@link MatchService#calculatePointsForTeams} (shared
+     * constants); ties break by goal difference, then goals scored, then name.
+     */
+    public StandingsDto getStandings() {
+        var finishedMatches = matchRepository.findAllByHomeScoreNotNullAndAwayScoreNotNull();
 
-        var comparatorByPoints = Comparator.comparingInt(Team::getPoints).reversed();
-        var teams = matches.stream()
-                .flatMap(match -> Stream.of(match.getHome(), match.getAway()))
-                .distinct()
-                .sorted(comparatorByPoints)
-                .collect(Collectors.toList());
+        var statsByTeamId = new HashMap<Long, TeamStats>();
+        for (var match : finishedMatches) {
+            statsByTeamId.computeIfAbsent(match.getHome().getId(), id -> new TeamStats())
+                    .addResult(match.getHomeScore(), match.getAwayScore());
+            statsByTeamId.computeIfAbsent(match.getAway().getId(), id -> new TeamStats())
+                    .addResult(match.getAwayScore(), match.getHomeScore());
+        }
 
-        var teamIds = teams.stream()
-                .map(Team::getId)
+        var sortedTeams = teamRepository.findAll().stream()
+                .sorted(tableOrder(statsByTeamId))
                 .toList();
-        List<Team> teamsWithoutMatches;
-        // TODO probably this if could be removed
-        if (teamIds.isEmpty())
-            teamsWithoutMatches = teamRepository.findAll();
-        else
-            teamsWithoutMatches = teamRepository.findAllByIdNotIn(teamIds);
-        teams.addAll(teamsWithoutMatches);
+        var rows = IntStream.range(0, sortedTeams.size())
+                .mapToObj(i -> toRow(sortedTeams.get(i), (short) (i + 1), statsByTeamId))
+                .toList();
 
-        return teams;
+        var afterQueue = finishedMatches.stream()
+                .map(Match::getQueue)
+                .max(Short::compare)
+                .orElse(null);
+        return new StandingsDto(afterQueue, rows);
+    }
+
+    private static Comparator<Team> tableOrder(Map<Long, TeamStats> statsByTeamId) {
+        Comparator<Team> byPoints = Comparator.comparingInt(team -> statsOf(team, statsByTeamId).points());
+        Comparator<Team> byGoalDifference = Comparator.comparingInt(team -> statsOf(team, statsByTeamId).goalDifference());
+        Comparator<Team> byGoalsFor = Comparator.comparingInt(team -> statsOf(team, statsByTeamId).goalsFor);
+        return byPoints.reversed()
+                .thenComparing(byGoalDifference.reversed())
+                .thenComparing(byGoalsFor.reversed())
+                .thenComparing(Team::getName);
+    }
+
+    private static StandingsDto.Row toRow(Team team, short place, Map<Long, TeamStats> statsByTeamId) {
+        var stats = statsOf(team, statsByTeamId);
+        return new StandingsDto.Row(team.getId(), team.getName(), team.getCity(), team.getShortCode(),
+                (short) stats.points(), place, (short) stats.played, (short) stats.wins, (short) stats.draws,
+                (short) stats.losses, (short) stats.goalsFor, (short) stats.goalsAgainst);
+    }
+
+    private static TeamStats statsOf(Team team, Map<Long, TeamStats> statsByTeamId) {
+        return statsByTeamId.getOrDefault(team.getId(), NO_MATCHES);
+    }
+
+    /** Per-team accumulator. Ints internally — narrowed to short only when building the DTO row. */
+    private static final class TeamStats {
+        private int played;
+        private int wins;
+        private int draws;
+        private int losses;
+        private int goalsFor;
+        private int goalsAgainst;
+
+        void addResult(short scored, short conceded) {
+            played++;
+            goalsFor += scored;
+            goalsAgainst += conceded;
+            if (scored > conceded)
+                wins++;
+            else if (scored < conceded)
+                losses++;
+            else
+                draws++;
+        }
+
+        int points() {
+            return wins * MatchService.POINTS_FOR_WIN_MATCH + draws * MatchService.POINTS_FOR_DRAW_MATCH;
+        }
+
+        int goalDifference() {
+            return goalsFor - goalsAgainst;
+        }
     }
 }
--- a/src/main/webapp/src/app/component/match-detail/match-detail.component.ts
+++ b/src/main/webapp/src/app/component/match-detail/match-detail.component.ts
@@ -2,7 +2,7 @@ import {Component, OnInit, computed, inject, signal, ChangeDetectionStrategy} fr
 import {ActivatedRoute, Router, RouterLink} from '@angular/router';
 import {forkJoin} from 'rxjs';
 import {Match} from '../../model/match';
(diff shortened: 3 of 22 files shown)


src/main/java/com/jamex/refereestaffer/service/TeamService.java
```diff
@@ -1,51 +1,110 @@
...
+
+    private static Comparator<Team> tableOrder(Map<Long, TeamStats> statsByTeamId) {
+        Comparator<Team> byPoints = Comparator.comparingInt(team -> statsOf(team, statsByTeamId).points());
+        Comparator<Team> byGoalDifference = Comparator.comparingInt(team -> statsOf(team, statsByTeamId).goalDifference());
+        Comparator<Team> byGoalsFor = Comparator.comparingInt(team -> statsOf(team, statsByTeamId).goalsFor);
+        return byPoints.reversed()
+                .thenComparing(byGoalDifference.reversed())
+                .thenComparing(byGoalsFor.reversed())
+                .thenComparing(Team::getName);
+    }
```
UnLow1, line 68: **Behaviour change vs the old endpoint:** the previous implementation sorted by points only, so equal-points teams kept incidental insertion order (and teams without matches were appended in repository order). This comparator makes ties deterministic with the conventional league chain — goal difference, then goals scored, then name. If the real Ekstraklasa tie-breaker rules (head-to-head first) ever matter here, this is the one place to change.

---

src/main/java/com/jamex/refereestaffer/service/TeamService.java
```diff
@@ -1,51 +1,110 @@
...
+                    .addResult(match.getAwayScore(), match.getHomeScore());
+        }
 
-        var teamIds = teams.stream()
-                .map(Team::getId)
+        var sortedTeams = teamRepository.findAll().stream()
+                .sorted(tableOrder(statsByTeamId))
+                .toList();
+        var rows = IntStream.range(0, sortedTeams.size())
+                .mapToObj(i -> toRow(sortedTeams.get(i), (short) (i + 1), statsByTeamId))
```
UnLow1, line 50: `place` here ranks **every** team 1..N (unplayed teams sort to the bottom), which is what the table screens need. Note this deliberately differs from the transient `Team.place` semantics used by the difficulty path (`MatchService.calculatePointsForTeams`), where `place == 0` means "unranked" and gates the top/bottom edge-match bonus. The two stay separate on purpose — unifying them is the `Team.place` modelling follow-up mentioned in RS-27/RS-72 territory (drop the transient fields from the entity once the staffer path computes its own ranking).

---

src/main/java/com/jamex/refereestaffer/model/dto/StandingsDto.java
```diff
@@ -0,0 +1,38 @@
...
+     * One table row. Carries the same team fields as {@link TeamDto} (so team pills
+     * render from a row directly) plus the season stats. `place` is 1-based table
+     * position — teams without a finished match sort to the bottom with zeroed stats.
+     */
+    public record Row(
+            Long id,
+            String name,
+            String city,
+            @JsonProperty("short") String shortCode,
+            short points,
```
UnLow1, line 29: Design decision: rows are a **flat superset of the `TeamDto` wire shape** (rather than nesting a `TeamDto`), so the frontend `Standing extends Team` and a row drops into every `Team`-typed slot (team pills, `teamsById` lookups) without mapping.

Side effect worth knowing: with this endpoint no longer using `TeamConverter`, the `points` field on `TeamDto` is now always `0` on the wire (`GET /api/teams` reads the transient entity field that nothing populates outside the old standings path). Removing `TeamDto.points` + the converter line would be a tidy follow-up, but it's API churn beyond this ticket so I left it.

---

src/main/webapp/src/app/component/match-detail/match-detail.component.ts
```diff
@@ -101,9 +101,11 @@ export class MatchDetailComponent implements OnInit {
...
       {label: 'Position', home: hp != null ? `#${hp}` : '—', away: ap != null ? `#${ap}` : '—', bar: false},
       {label: 'Points', home: h.points ?? 0, away: a.points ?? 0, bar: true},
-      // W / D / L / GF / GA need a backend stats endpoint that doesn't exist yet. For
-      // now they're omitted; placeholder note in the template tells the reader why the
-      // panel feels short.
+      {label: 'Wins', home: h.wins, away: a.wins, bar: true},
+      {label: 'Draws', home: h.draws, away: a.draws, bar: true},
+      {label: 'Losses', home: h.losses, away: a.losses, bar: true},
+      {label: 'Goals for', home: h.goalsFor, away: a.goalsFor, bar: true},
+      {label: 'Goals against', home: h.goalsAgainst, away: a.goalsAgainst, bar: true},
```
UnLow1, line 109: Slight scope extension beyond the ticket's "standings + team-list" — but the old comment here (and the placeholder note in the template) said these rows were blocked on exactly the endpoint this PR adds, and the data now arrives on a response this component already loads. Adding the five rows costs nothing extra on the wire; leaving the "endpoint doesn't exist yet" note would have been stale.

---

UnLow1 (commented):
Self-review notes on the discussable spots — tie-breaker semantics, the two `place` semantics, the flat DTO shape (and the now-dead `TeamDto.points`), and the small match-detail scope extension.

---

src/main/java/com/jamex/refereestaffer/service/TeamService.java
UnLow1: Resolution: keeping the GD → GF → name chain. The old equal-points ordering was an accident of stream insertion order (not a behaviour anyone chose), the new chain is the convention most league tables use, and it's covered by a dedicated spec (`"should break point ties by goal difference, then goals scored, then name"`). Head-to-head can be layered in here later if it's ever needed — one comparator to touch.

---

src/main/java/com/jamex/refereestaffer/service/TeamService.java
UnLow1: Resolution: no code change. The frontend consumers of this endpoint (standings table, staffer/match-detail place lookups) all want "every team ranked", and the difficulty path keeps its own 0-means-unranked guard untouched — its specs still pass unmodified. Unifying the two rankings belongs to the entity-modelling refactor (RS-72 area), not this ticket.

---

src/main/java/com/jamex/refereestaffer/model/dto/StandingsDto.java
UnLow1: Resolution: leaving `TeamDto.points` in place. Double-checked it's not a regression from this PR — `GET /api/teams` / `/byIds` / `/{id}` already returned `points: 0` before (the transient field was only populated on the old standings path), and no frontend code reads `points` off those responses. Removing the field is a clean follow-up (TeamDto + TeamConverter + its spec + the frontend `Team` interface), best done together with the `Team` transient-fields refactor.

---

src/main/webapp/src/app/component/match-detail/match-detail.component.ts
UnLow1: Resolution: keeping the comparison rows. The change is confined to this component (five entries in an already-generic `compareRows` list plus deleting the placeholder note), adds no extra request, and directly closes the "blocked on a season-stats endpoint" TODO this PR unblocks. If it turns out unwanted, reverting is a five-line deletion.

---

UnLow1:
Status analysis of this PR with respect to `Team.place`, written down so it doesn't have to be redone. Context: RS-74 (#88) has just merged, RS-75 is waiting in #89.

## What this PR does with `place` — and what it leaves behind

It delivers exactly the variant the ticket called "computed in a query, not on the entity" — but **only for the standings endpoint**. `TeamService.getStandings()` returns `StandingsDto`, accumulating stats in a `HashMap<Long, TeamStats>` keyed by `teamId`, sorting by points → goal difference → goals for → name, and assigning `place = index + 1` straight into the DTO row. No entity mutation, the `MatchService` dependency is gone, `findAllByIdNotIn` is removed from the repository.

What it does not do: `MatchService.calculatePointsForTeams` stays untouched (as the description states), and `Team` keeps `@Transient points` and `@Transient place`. Verified in the branch's code, not just in the description.

**Net state once this merges: two parallel league-table implementations.** The new read model serves `/api/teams/standings`; the old entity-mutating path remains for the staffer (`getMatchesToAssignInQueue`) and difficulty (`computeDifficultyBreakdown`) flows. So RS-27 narrows `place` from three consumers to two rather than closing it out. The third step — pointing the algorithm path at the same read model and then deleting both transients from the entity — is currently covered by no ticket.

## Interaction with RS-75 (#89) — matters for ordering

While working on RS-74 I proved empirically that disabling OSIV silently breaks `GET /api/matches/{id}/difficulty`: `computeDifficultyBreakdown` loads the match in one repository call but computes points/places on the entities returned by a second one, so without OSIV those are different instances and the endpoint returns HTTP 200 with `pointsDiff = 0` and no edge bonuses. Full write-up and reproduction in [#89](https://github.com/UnLow1/Referee-Staffer/pull/89#issuecomment-5160378327).

This PR does **not** fix that (the difficulty path still reads the transients), but it makes the fix substantially simpler: the new `StandingsDto` is keyed by `teamId` and is therefore inherently immune to the persistence-context split. If this lands first, the RS-75 fix reduces to consuming the existing read model in `computeDifficultyBreakdown` instead of bolting on an ad-hoc map — and only then can `place`/`points` leave the entity. Hence the suggested order: **#83 → #89**.

## Correction: this PR does not close RS-84

I had assumed backend standings would resolve RS-84 as a side effect. Not true as written — this PR deliberately keeps the top/bottom zone logic client-side as pure presentation, and `const NUMBER_OF_EDGE_TEAMS = 3` remains hardcoded in all four components (`dashboard`, `staffer`, `standings`, `match-detail`). The improvement is real but partial: `place` now comes from the backend instead of being re-derived from the array index. The constant itself stays duplicated, so RS-84 remains open.

## Rebase state — smaller than the "dirty" flag suggests

Trial rebase on a throwaway local branch (nothing pushed):

- the **backend rebases 100% cleanly** — zero conflicts, despite DTO records (#99), the Java 26 bump and RS-74 (#88) landing in the meantime;
- after `mvn clean`, **147/147 backend tests pass** on the rebased tree;
- only **3 frontend files conflict**: `standings.component.ts`, `standings.component.html`, `team-list.component.html` — one hunk each, import-line collisions originating from RS-85 (#94) and RS-79 (#72).

One verification trap worth recording: the first test run showed 2 failures that turned out to be stale classes in `target/` from a previous build of a different branch (Surefire ran specs that do not exist in this tree). The 147 figure is from a run after `mvn clean` — clean between branches when repeating this check.

**The real risk is not the rebase, it's Angular.** Master has moved to Angular 22 + TypeScript 6 (RS-79) while this PR was written against Angular 21. The textual conflicts are mechanical, but after rebasing the frontend needs a fresh `npm ci && npm run lint && npm test && npm run build` — compatibility with the new build has to be verified, not assumed.

🤖 Generated with Claude Code

---

UnLow1:
**Correction to my previous comment.** There I wrote that if this PR lands first, the RS-75 fix "reduces to consuming the existing read model in `computeDifficultyBreakdown`". That was wrong and I'm withdrawing it.

Comparing the two league-table implementations, they produce **different places**, so pointing the algorithm at this PR's read model is a behaviour change rather than a refactor:

| | old (`calculatePointsForTeams`) | new (`getStandings`, this PR) |
|---|---|---|
| ranked teams | only those with a finished match | all teams from `findAll()` |
| sorting | points only | points → goal difference → goals for → name |
| ties | stream encounter order | explicit tie-break chain |
| team with no matches | unranked (`place = null` after RS-74) | real place at the bottom, zeroed stats |

Match hardness would change as a result — an unplayed team could become a bottom-of-table fixture, and ties would resolve differently.

This does not detract from this PR; the new read model is the better model, and it is keyed by `teamId`, which is exactly why it is the right long-term target. It just means the swap deserves its own PR with hardness assertions instead of riding along inside an OSIV config change. Tracked as [RS-99](https://referee-staffer.youtrack.cloud/issue/RS-99).

Worth recording while it is in view: RS-99 also captures an existing inconsistency this PR's model would incidentally fix. `computeEdgeMatchParts` compares place against `teamRepository.count()` (all teams) while places are assigned only among teams with finished matches — with 16 teams and 6 played, the bottom-zone condition is `place > 13` against a highest existing place of 6, so the bottom-of-table bonus never fires early in a season. Once the ranked population and the divisor are the same set, it starts working.

Revised order for the cluster: **this PR → RS-84 → RS-75 (#89, behaviour-preserving fix) → RS-99 → RS-70**. Everything else in my previous comment (clean backend rebase, 147/147 after `mvn clean`, three one-hunk frontend conflicts, Angular 22 revalidation as the real risk) stands.

🤖 Generated with Claude Code

---

PR #84, opened 2026-07-12
RS-57: Skip referees with another match on the same day when staffing

UnLow1:
## Ticket

[RS-57 — Przy tworzeniu obsady sprawdzać kolizję w datach](https://referee-staffer.youtrack.cloud/issue/RS-57)

By design a referee can't get two matches in the same queue, and the staffer already enforces that (`findAllWithNoMatchInQueue` + the in-run `busy` flag). But a match can be rescheduled from its queue's usual weekend onto another date — so a referee who already officiates a rescheduled match from *another* queue could be assigned a second match on the very same day. The staffer needs to treat "already has a match that day" as a hard exclusion, same as vacations.

## Plan

1. **Repository**: add a day-bounded lookup to `MatchRepository` — a derived query over `[dayStart, nextDayStart)` plus a `findAllByRefereeInAndDateOnDay(referees, dateTime)` default method, following the existing `VacationRepository` pattern of wrapping a `LocalDate`-window query behind a `LocalDateTime` convenience method.
2. **Service**: inject `MatchRepository` into `StafferService`; in `assignRefereesToMatches`, for each match compute the set of candidates who already have any match on that calendar day (any queue) and filter them out alongside the existing `busy` and vacation filters. Guard the empty candidate list so no query runs with an empty `IN`.
3. **Tests**: unit features in `StafferServiceSpec` (conflicted referee skipped in favour of a lower-rated one; `StafferException` when every candidate conflicts; no same-day query for an empty pool) plus an end-to-end case in `StafferIntegrationSpec` against real H2, which also exercises the derived query and the default method's day-window arithmetic that mocks can't reach.
4. **Risks considered**: entity identity in the filter (same persistence context → same instances, no `equals` override needed); Hibernate's AUTO flush making in-run assignments visible to the per-match query (consistent with the `busy` flag, no conflict); one extra query per staffed match (~8 per run, same cost profile as the existing per-match vacation query).

## Changes

- `MatchRepository`: new derived query `findAllByRefereeInAndDateGreaterThanEqualAndDateLessThan` + `findAllByRefereeInAndDateOnDay` default method. The window is half-open (`>= dayStart`, `< dayStart.plusDays(1)`), so any time on the day matches and midnight of the next day doesn't.
- `StafferService`: new `findRefereesWithMatchOnDay` helper feeding a third filter in the candidate stream. Empty referee pool short-circuits to `Set.of()` — the existing `StafferException` path then reports the failure as before.
- `StafferServiceSpec`: two new features for the collision rule, `0 *` interaction documenting the empty-pool short-circuit, and stubs/dates added to existing features (match `date` is `nullable = false` in the entity, so the added dates also mirror the real invariant).
- `StafferIntegrationSpec`: new feature where the higher-potential referee already has a queue-1 match rescheduled onto the staffed day and the lower-potential one gets the assignment; annotated the spec `@Execution(SAME_THREAD)` because its features share the context-cached H2 instance that `setup()` wipes — Spock's parallel mode would let them race.

Deliberately out of scope: the manual referee change via `PUT /api/matches` (plain repository save, no validation layer today) and the related "minimum rest between matchdays" rule — both are separate backlog items.

## Testing

- `mvn -DskipFrontend=true test` — 133 tests green (12 in `StafferServiceSpec`, 2 in `StafferIntegrationSpec`).
- The integration case fails without the fix: with `EXPERIENCE_MULTIPLIER = 0.01` from `data.sql` and equal (fallback) grades, the same-day-busy referee wins on potential unless excluded.

https://claude.ai/code/session_01Hdwc32566RqCF1eL14AdWc

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hdwc32566RqCF1eL14AdWc)_

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/repository/MatchRepository.java
+++ b/src/main/java/com/jamex/refereestaffer/repository/MatchRepository.java
@@ -5,6 +5,7 @@ import com.jamex.refereestaffer.model.entity.Referee;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
+import java.time.LocalDateTime;
 import java.util.Collection;
 import java.util.List;
 
@@ -16,4 +17,12 @@ public interface MatchRepository extends JpaRepository<Match, Long> {
     List<Match> findAllByQueueAndRefereeIsNull(Short queue);
 
     List<Match> findAllByHomeScoreNotNullAndAwayScoreNotNull();
+
+    List<Match> findAllByRefereeInAndDateGreaterThanEqualAndDateLessThan(Collection<Referee> referees,
+                                                                         LocalDateTime from, LocalDateTime to);
+
+    default List<Match> findAllByRefereeInAndDateOnDay(Collection<Referee> referees, LocalDateTime dateTime) {
+        var dayStart = dateTime.toLocalDate().atStartOfDay();
+        return findAllByRefereeInAndDateGreaterThanEqualAndDateLessThan(referees, dayStart, dayStart.plusDays(1));
+    }
 }
--- a/src/main/java/com/jamex/refereestaffer/service/StafferService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/StafferService.java
@@ -9,18 +9,21 @@ import com.jamex.refereestaffer.model.entity.Team;
 import com.jamex.refereestaffer.model.entity.Vacation;
 import com.jamex.refereestaffer.model.exception.StafferException;
 import com.jamex.refereestaffer.repository.ConfigurationRepository;
+import com.jamex.refereestaffer.repository.MatchRepository;
 import com.jamex.refereestaffer.repository.VacationRepository;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.time.LocalDateTime;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 @Service
@@ -30,14 +33,17 @@ public class StafferService {
 
     private final ConfigurationRepository configurationRepository;
     private final VacationRepository vacationRepository;
+    private final MatchRepository matchRepository;
     private final MatchConverter matchConverter;
     private final MatchService matchService;
     private final RefereeService refereeService;
 
     public StafferService(ConfigurationRepository configurationRepository, VacationRepository vacationRepository,
-                          MatchConverter matchConverter, MatchService matchService, RefereeService refereeService) {
+                          MatchRepository matchRepository, MatchConverter matchConverter, MatchService matchService,
+                          RefereeService refereeService) {
         this.configurationRepository = configurationRepository;
         this.vacationRepository = vacationRepository;
+        this.matchRepository = matchRepository;
         this.matchConverter = matchConverter;
         this.matchService = matchService;
         this.refereeService = refereeService;
@@ -66,9 +72,12 @@ public class StafferService {
                     .map(Vacation::getReferee)
                     .toList();
 
+            var refereesWithMatchOnSameDay = findRefereesWithMatchOnDay(referees, match.getDate());
+
             var availableReferees = referees.stream()
                     .filter(ref -> !ref.isBusy())
                     .filter(ref -> !refereesWithVacations.contains(ref))
+                    .filter(ref -> !refereesWithMatchOnSameDay.contains(ref))
                     .toList();
 
             for (var referee : availableReferees) {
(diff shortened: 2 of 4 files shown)


src/main/java/com/jamex/refereestaffer/repository/MatchRepository.java
```diff
@@ -16,4 +17,12 @@ public interface MatchRepository extends JpaRepository<Match, Long> {
     List<Match> findAllByQueueAndRefereeIsNull(Short queue);
 
     List<Match> findAllByHomeScoreNotNullAndAwayScoreNotNull();
+
+    List<Match> findAllByRefereeInAndDateGreaterThanEqualAndDateLessThan(Collection<Referee> referees,
+                                                                         LocalDateTime from, LocalDateTime to);
+
+    default List<Match> findAllByRefereeInAndDateOnDay(Collection<Referee> referees, LocalDateTime dateTime) {
```
UnLow1, line 24: Reviewer context: this mirrors the `VacationRepository` idiom (derived `LocalDate`-window query behind a `LocalDateTime` convenience default method). One consequence worth knowing: Spock mocks intercept default methods too, so the unit specs stub this method directly and the day-window arithmetic in its body is only actually executed by `StafferIntegrationSpec`. The alternative — computing `dayStart`/`dayStart.plusDays(1)` in `StafferService` and calling the derived query directly — would make the window unit-testable, at the cost of breaking the repo's established pattern. Went with consistency; happy to flip if you prefer the other trade-off.

---

src/main/java/com/jamex/refereestaffer/service/StafferService.java
```diff
@@ -91,6 +100,21 @@ private void assignRefereesToMatches(List<Referee> referees, List<Match> matches
...
+     * Referees that already officiate another match on the same calendar day. The queue-level
+     * uniqueness check (getAvailableRefereesForQueue + busy) does not cover this: a match from a
+     * different queue can be rescheduled onto this day, and one referee must never have two
+     * matches on one day (RS-57).
+     */
+    private Set<Referee> findRefereesWithMatchOnDay(List<Referee> referees, LocalDateTime date) {
+        if (referees.isEmpty()) {
+            return Set.of();
+        }
+        return matchRepository.findAllByRefereeInAndDateOnDay(referees, date).stream()
```
UnLow1, line 113: Two subtleties here:

1. The membership check downstream relies on entity identity (no `equals`/`hashCode` on `Referee`). That's safe because everything runs inside the `@Transactional` staffing flow — the referees returned by this query and the candidates from `getAvailableRefereesForQueue` come from the same persistence context, so Hibernate hands back the same instances.
2. This runs once per staffed match (like the vacation lookup above), and Hibernate's AUTO flush means assignments made earlier in this loop are visible to it — redundant with the `busy` flag for same-queue matches, but never contradictory. Batching per distinct day would save a few queries; at ~8 matches per queue it didn't seem worth the extra code.

---

src/test/groovy/com/jamex/refereestaffer/integration/StafferIntegrationSpec.groovy
```diff
@@ -18,7 +20,11 @@ import java.time.LocalDateTime
  * The unit tests in {@link com.jamex.refereestaffer.service.StafferServiceSpec} cover the
  * algorithm with mocks — this one's job is to prove that referee assignment actually
  * persists to the database, which is the part dependency-injected mocks can never verify.
+ *
+ * <p>SAME_THREAD because all features share the one H2 instance behind the cached Spring
+ * context and setup() wipes it — Spock's parallel mode would let features race on that state.
  */
+@Execution(ExecutionMode.SAME_THREAD)
```
UnLow1, line 27: This annotation becomes necessary only now that the spec has a second feature: `SpockConfig.groovy` enables parallel execution with default (concurrent) mode, and both features share the one H2 database behind the cached Spring context while `setup()` calls `deleteAll()` on it. With a single feature the race couldn't happen, so the original spec didn't need it. Same idiom the `@WebMvcTest` controller specs already use for their shared `@SpringBean` mocks.

---

UnLow1 (commented):

---

src/main/java/com/jamex/refereestaffer/repository/MatchRepository.java
UnLow1: Acted on the real gap behind this: the window's boundaries had no test at all. Added an integration case in cb591c7 where the candidate's other matches sit at 23:00 the day before and 00:00 the day after — both must stay outside the `[dayStart, nextDayStart)` exclusion window, so any off-by-one in the default method now fails a test. Keeping the default-method placement itself (consistency with `VacationRepository`); 134 tests green.

---

src/main/java/com/jamex/refereestaffer/service/StafferService.java
UnLow1: Keeping as is. Point 1 is an invariant of the current `@Transactional` design, not a latent bug — if `staffReferees` ever stops being transactional, the integration test (which re-reads persisted state) is positioned to catch fallout. Point 2 is a deliberate simplicity-over-batching call: the per-match query matches the existing vacation-lookup cost profile, and matches within one staffed queue usually share a day anyway, so the win from grouping would be a handful of queries per run.

---

src/test/groovy/com/jamex/refereestaffer/integration/StafferIntegrationSpec.groovy
UnLow1: No change needed — the annotation is required from the second feature onward and follows the same convention the controller specs already use. Noting for posterity: if integration specs multiply, a shared base class (or `@ResourceLock` on the H2 state) would be the cleaner scaling move.

---

PR #85, opened 2026-07-13
RS-71: Move match reference resolution from MatchConverter to MatchService

UnLow1:
## Ticket

[RS-71: MatchConverter bez repozytoriów — rozwiązywanie referencji w serwisie](https://referee-staffer.youtrack.cloud/issue/RS-71)

`MatchConverter` injected three repositories (`TeamRepository`, `RefereeRepository`, `GradeRepository`) and queried the database on every DTO → entity conversion. The conversion layer should be pure mapping (SRP) — and because `BaseConverter.convertFromDtos` maps element by element, the bulk `PUT /api/matches` endpoint issued 2–4 hidden queries **per match** (a classic N+1). The ticket asks to move id → entity resolution to the service layer and hand the converter ready entities.

## Plan

1. **`MatchConverter`** — drop the repository dependencies; keep `convertFromEntity`/`convertFromEntities`; change `convertFromDto` to take already-resolved `Team`/`Referee`/`Grade` entities. Since the dto → entity half now needs context, the class stops implementing `BaseConverter` (nothing uses that interface polymorphically; the other four converters keep it).
2. **`MatchService`** — add `saveMatch(MatchDto)` and `updateMatches(List<MatchDto>)`, with a shared private resolution step that loads all referenced entities in bulk (`findAllById`, one query per repository) and preserves the existing semantics: missing team → `TeamNotFoundException` (404), missing/null referee or grade id → `null` on the entity.
3. **`MatchController`** — delegate `POST /api/matches`, `PUT /api/matches/{id}` and `PUT /api/matches` to the service; GET endpoints keep using the repository + entity → DTO conversion as before.
4. **Tests** — rewrite `MatchConverterSpec` as a pure spec (no mocks), move the `TeamNotFoundException` case to `MatchServiceSpec`, cover the new resolution logic (bulk queries, optional refs, unknown ids), update `MatchControllerSpec` stubs to the service methods. Run `mvn -DskipFrontend=true test`.

Risks considered: behaviour parity of the not-found semantics (kept exactly), and Spring context wiring (`StafferService` uses only the entity → DTO direction, so it is unaffected; no dependency cycle is introduced — `MatchService` → `MatchConverter` is one-way).

## Changes

- `MatchConverter` is now a pure, dependency-free mapper. `convertFromDto(MatchDto, Team, Team, Referee, Grade)` just builds the entity. It no longer implements `BaseConverter`, so it declares its own `convertFromEntities` (same 3-line stream as the interface default).
- `MatchService` gained `saveMatch`/`updateMatches` plus a private `resolveAndConvert` that collects all team/referee/grade ids from the DTOs and loads each entity type with **one** `findAllById` — the bulk update endpoint went from ~2–4 queries per match to 3 queries total. Single create/update reuses the same path with a one-element list.
- Missing-reference semantics are unchanged: unknown team id throws `TeamNotFoundException` (checked home first, then away, per DTO — same order as the old per-id lookups), unknown or null referee/grade ids resolve to `null`.
- `MatchController` POST/PUT endpoints are thin delegates to `MatchService` now; both single-match endpoints use the same `saveMatch` (their bodies were previously identical).
- Out of scope, left as-is: `GradeConverter` and `VacationConverter` have the same repository-in-converter smell — worth a follow-up ticket if desired.

## Testing

- `MatchConverterSpec` rewritten as a pure spec — no repository mocks left.
- `MatchServiceSpec`: 5 new features covering resolution — happy path with all references, optional refs absent (verifies referee/grade queries are skipped entirely), unknown referee/grade ids resolving to null, `TeamNotFoundException` for a missing home/away team, and bulk update asserting exactly one `findAllById` per repository.
- `MatchControllerSpec`: POST/PUT features now stub `MatchService` instead of converter + repository.
- Full backend suite: `mvn -DskipFrontend=true test` → **136 tests, 0 failures** (on JDK 25). Frontend untouched.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01R61MigTZJQ3QcyLphf3crD)_

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/model/converter/MatchConverter.java
+++ b/src/main/java/com/jamex/refereestaffer/model/converter/MatchConverter.java
@@ -4,28 +4,23 @@ import com.jamex.refereestaffer.model.dto.MatchDto;
 import com.jamex.refereestaffer.model.entity.Grade;
 import com.jamex.refereestaffer.model.entity.Match;
 import com.jamex.refereestaffer.model.entity.Referee;
-import com.jamex.refereestaffer.model.exception.TeamNotFoundException;
-import com.jamex.refereestaffer.repository.GradeRepository;
-import com.jamex.refereestaffer.repository.RefereeRepository;
-import com.jamex.refereestaffer.repository.TeamRepository;
+import com.jamex.refereestaffer.model.entity.Team;
 import org.springframework.stereotype.Component;
 
+import java.util.Collection;
 import java.util.Optional;
+import java.util.stream.StreamSupport;
 
+/**
+ * Pure mapping between {@link Match} and {@link MatchDto} — no repository access.
+ * Resolving ids to entities (teams, referee, grade) is the service layer's job
+ * ({@link com.jamex.refereestaffer.service.MatchService}), so the dto → entity
+ * direction takes already-resolved entities instead of implementing
+ * {@link BaseConverter#convertFromDto(Object)}.
+ */
 @Component
-public class MatchConverter implements BaseConverter<Match, MatchDto> {
+public class MatchConverter {
 
-    private final TeamRepository teamRepository;
-    private final RefereeRepository refereeRepository;
-    private final GradeRepository gradeRepository;
-
-    public MatchConverter(TeamRepository teamRepository, RefereeRepository refereeRepository, GradeRepository gradeRepository) {
-        this.teamRepository = teamRepository;
-        this.refereeRepository = refereeRepository;
-        this.gradeRepository = gradeRepository;
-    }
-
-    @Override
     public MatchDto convertFromEntity(Match entity) {
         var referee = Optional.ofNullable(entity.getReferee())
                 .map(Referee::getId);
@@ -45,25 +40,21 @@ public class MatchConverter implements BaseConverter<Match, MatchDto> {
                 entity.getHardnessLvl());
     }
 
-    @Override
-    public Match convertFromDto(MatchDto dto) {
-        var homeTeam = teamRepository.findById(dto.getHomeTeamId())
-                .orElseThrow(() -> new TeamNotFoundException(dto.getHomeTeamId()));
-        var awayTeam = teamRepository.findById(dto.getAwayTeamId())
-                .orElseThrow(() -> new TeamNotFoundException(dto.getAwayTeamId()));
-        var referee = Optional.ofNullable(dto.getRefereeId())
-                .flatMap(refereeRepository::findById);
-        var grade = Optional.ofNullable(dto.getGradeId())
-                .flatMap(gradeRepository::findById);
+    public Collection<MatchDto> convertFromEntities(final Iterable<Match> entities) {
+        return StreamSupport.stream(entities.spliterator(), false)
+                .map(this::convertFromEntity)
+                .toList();
+    }
 
+    public Match convertFromDto(MatchDto dto, Team home, Team away, Referee referee, Grade grade) {
         return new Match(
                 dto.getId(),
                 dto.getQueue(),
-                homeTeam,
-                awayTeam,
+                home,
+                away,
                 dto.getDate(),
-                referee.orElse(null),
-                grade.orElse(null),
+                referee,
+                grade,
                 dto.getHomeScore(),
                 dto.getAwayScore(),
                 0.0);
--- a/src/main/java/com/jamex/refereestaffer/service/MatchService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/MatchService.java
@@ -1,13 +1,19 @@
 package com.jamex.refereestaffer.service;
 
+import com.jamex.refereestaffer.model.converter.MatchConverter;
 import com.jamex.refereestaffer.model.dto.DifficultyBreakdownDto;
+import com.jamex.refereestaffer.model.dto.MatchDto;
 import com.jamex.refereestaffer.model.entity.ConfigName;
+import com.jamex.refereestaffer.model.entity.Grade;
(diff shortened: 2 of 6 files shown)


src/main/java/com/jamex/refereestaffer/model/converter/MatchConverter.java
```diff
@@ -4,28 +4,23 @@
...
-    private final TeamRepository teamRepository;
-    private final RefereeRepository refereeRepository;
-    private final GradeRepository gradeRepository;
-
-    public MatchConverter(TeamRepository teamRepository, RefereeRepository refereeRepository, GradeRepository gradeRepository) {
-        this.teamRepository = teamRepository;
-        this.refereeRepository = refereeRepository;
-        this.gradeRepository = gradeRepository;
-    }
+public class MatchConverter {
```
UnLow1, line 22: Design decision worth flagging: `MatchConverter` deliberately stops implementing `BaseConverter` instead of the interface being split into entity→dto / dto→entity halves. Nothing in the codebase uses `BaseConverter` polymorphically (each converter is injected as its concrete type), so an interface-segregation refactor would have added abstraction with no consumer. The cost is the duplicated 3-line `convertFromEntities` below — if a second converter ever needs the same treatment (`GradeConverter` and `VacationConverter` have the same repository-in-converter smell), splitting the interface would be the right move then.

---

src/main/java/com/jamex/refereestaffer/service/MatchService.java
```diff
@@ -31,13 +40,72 @@ public class MatchService {
...
+        return matchesDtos.stream()
+                .map(dto -> matchConverter.convertFromDto(dto,
+                        requireTeam(teams, dto.getHomeTeamId()),
+                        requireTeam(teams, dto.getAwayTeamId()),
+                        resolveOptional(referees, dto.getRefereeId()),
+                        resolveOptional(grades, dto.getGradeId())))
+                .toList();
+    }
+
+    private static <E> E resolveOptional(Map<Long, E> entitiesById, Long id) {
```
UnLow1, line 92: Behaviour-parity note: an **unknown** (non-null) referee or grade id silently resolves to `null` here instead of failing with a 404. That mirrors the pre-refactor code exactly (`Optional.ofNullable(id).flatMap(repo::findById)` swallowed missing ids the same way), so this PR keeps it — but it is arguably a latent bug: a client sending `refereeId: 999` gets a match saved with no referee and no error. If we ever want to tighten this, it's a one-line change to throw `RefereeNotFoundException`/`GradeNotFoundException` here — kept out of this PR to stay a pure refactor.

---

src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
```diff
@@ -18,9 +22,139 @@ class MatchServiceSpec extends Specification {
...
     ConfigurationRepository configurationRepository = Mock()
     TeamRepository teamRepository = Mock()
+    RefereeRepository refereeRepository = Mock()
 
     def setup() {
-        matchService = new MatchService(matchRepository, gradeRepository, configurationRepository, teamRepository)
+        // The converter is a pure mapper since RS-71, so the real one is used instead of a mock —
+        // these features then cover the whole resolve-references + convert path end to end.
+        matchService = new MatchService(matchRepository, gradeRepository, configurationRepository,
+                teamRepository, refereeRepository, new MatchConverter())
```
UnLow1, line 31: Deliberate deviation from the spec's mock-everything style: the real `MatchConverter` is injected instead of a `Mock()`. Since the converter is now a pure mapper with no dependencies, mocking it would only re-assert the stubbing; using the real one makes these features cover the full resolve→convert path (e.g. the `match.referee == null` assertions actually exercise the converter). The repositories stay mocked, so the specs remain fast and DB-free.

---

UnLow1 (commented):
Self-review notes on discussable spots — details inline.

---

src/main/java/com/jamex/refereestaffer/model/converter/MatchConverter.java
UnLow1: Resolved as: keeping the code as-is. Dropping the interface on this one converter is the smallest change that matches the ticket's scope; an interface split (`EntityToDtoConverter` / `BaseConverter`) only pays off once a second converter loses its dto→entity repository access. `GradeConverter`/`VacationConverter` are candidates for a follow-up ticket — noted in the PR description under "out of scope".

---

src/main/java/com/jamex/refereestaffer/service/MatchService.java
UnLow1: Resolved as: keeping the silent-null semantics in this PR. Changing unknown `refereeId`/`gradeId` to a 404 would alter the public API behaviour inside a refactor ticket, and the frontend drawers currently only send ids picked from loaded lists, so nothing exercises the strict path today. Tightening this (throw `RefereeNotFoundException`/`GradeNotFoundException` in `resolveOptional`-callers) is a good candidate for a small follow-up ticket in YouTrack.

---

src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
UnLow1: Resolved as: keeping the real converter. A mocked converter would force every feature to stub `convertFromDto(...)` return values, and the assertions on the saved `Match` (resolved home/away/referee/grade) would then only prove the stubbing, not the behaviour. The unit under test is effectively "service + pure mapper", which is stable and dependency-free — same trade-off the specs already accept for other pure helpers.

---

PR #86, opened 2026-07-13
RS-72: Replace Referee.busy transient with local assigned-ids set in StafferService

UnLow1:
## Ticket

[RS-72](https://referee-staffer.youtrack.cloud/issue/RS-72) — *Referee.busy → Set&lt;Long&gt; w StafferService (stan procesu poza encją)*

The `Referee` JPA entity carried a `@Transient boolean busy` field that existed only to mark a referee as "already assigned" during a single staffing run. That mixes transient process state of the staffing algorithm into the domain model (an anti-pattern flagged by a long-standing `TODO` in `StafferService`). The goal was a pure refactor: move that bookkeeping into `StafferService` itself and remove the field from the entity, with no behavior change.

## Plan

1. In `StafferService.assignRefereesToMatches`, introduce a local `Set<Long> assignedRefereeIds` scoped to one staffing run; replace the `!ref.isBusy()` filter with `!assignedRefereeIds.contains(ref.getId())` and record `chosenReferee.getId()` after each assignment; drop the `TODO` and the `setBusy(true)` call.
2. Remove `busy` from the `Referee` entity: field, `isBusy()`/`setBusy()`, the all-args constructor parameter, and the builder field/method.
3. Update the only other caller of the all-args constructor, `RefereeConverter.convertFromDto`, to drop the trailing `false` argument.
4. Adjust `StafferServiceSpec` setups: referees in the two multi-referee features now need ids (deduplication is id-based, and the entity intentionally has no id setter, so map coercion can't provide one) — switch those setups to `Referee.builder()` with explicit ids. The ticket explicitly anticipated such minor setup fixes.
5. Run `mvn -DskipFrontend=true test` and confirm the whole backend suite is green.

## Changes

- `StafferService`: `assignRefereesToMatches` now tracks assigned referees in a local `HashSet<Long>` created at the start of each run, so the "busy" state lives and dies with a single `staffReferees` call — same lifecycle the transient field effectively had (entities were reloaded fresh each transaction), but now explicit and confined to the service.
- `Referee`: the `busy` field is gone from the entity, its all-args constructor, and the nested `Builder`.
- `RefereeConverter.convertFromDto`: constructor call updated accordingly.
- `StafferServiceSpec`: referee fixtures in "should assign referees to matches in queue" and "should not assign referees to matches if referee has vacation" are built with `Referee.builder()` and explicit ids (`1L`…`6L`). Without ids, all fixtures would share a `null` id and the id-based dedup would wrongly exclude every remaining referee after the first assignment. Production code always has DB-generated ids, so this is a test-fixture concern only.

Design decision: deduplication is by id (`Set<Long>`), as specified in the ticket, rather than by entity instance. Within one run the referee list comes from a single repository call, so both would work, but id-based tracking doesn't depend on instance identity or on `Referee` not overriding `equals`.

## Testing

- `mvn -DskipFrontend=true test` — all 130 Spock specs pass, including `StafferServiceSpec` (10 features, assignment + vacation + no-referee paths) and `StafferIntegrationSpec` (real Spring context + H2, verifies assignment persists).
- No frontend changes, so the frontend suite is untouched.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01M5y5wB8x23ReggCU4bMb6R)_

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/service/StafferService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/StafferService.java
@@ -18,6 +18,7 @@ import org.springframework.transaction.annotation.Transactional;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -58,6 +59,7 @@ public class StafferService {
     }
 
     private void assignRefereesToMatches(List<Referee> referees, List<Match> matches, Map<ConfigName, Double> config) {
+        var assignedRefereeIds = new HashSet<Long>();
         for (var match : matches) {
             var refereesPotentialLvlMap = new HashMap<Referee, Double>();
             var vacations = vacationRepository.findAllByStartDateIsLessThanEqualAndEndDateIsGreaterThanEqual(match.getDate());
@@ -67,7 +69,7 @@ public class StafferService {
                     .toList();
 
             var availableReferees = referees.stream()
-                    .filter(ref -> !ref.isBusy())
+                    .filter(ref -> !assignedRefereeIds.contains(ref.getId()))
                     .filter(ref -> !refereesWithVacations.contains(ref))
                     .toList();
 
@@ -84,8 +86,7 @@ public class StafferService {
             var chosenReferee = sortedRefereesPotentialLvlMap.keySet().stream()
                     .findFirst()
                     .orElseThrow(StafferException::new);
-            // TODO separate list for assigned referees id - get rid off field busy
-            chosenReferee.setBusy(true);
+            assignedRefereeIds.add(chosenReferee.getId());
 
             match.setReferee(chosenReferee);
         }
--- a/src/test/groovy/com/jamex/refereestaffer/service/StafferServiceSpec.groovy
+++ b/src/test/groovy/com/jamex/refereestaffer/service/StafferServiceSpec.groovy
@@ -37,9 +37,21 @@ class StafferServiceSpec extends Specification {
                 .build()
         // After RefereeService.calculateStats, averageGrade is always non-null — the
         // no-grades fallback (DEFAULT_GRADE = 8.3) is applied there. Setting it
-        // explicitly here mirrors the real invariant.
-        def ref1 = [averageGrade: 8.1d, teamsRefereed: Map.of(team1, (short) 1, team2, (short) 1), numberOfMatchesInRound: 7] as Referee
-        def ref2 = [averageGrade: RefereeService.DEFAULT_GRADE, experience: 100, teamsRefereed: [:], numberOfMatchesInRound: 0] as Referee
+        // explicitly here mirrors the real invariant. Ids are needed because the
+        // staffer deduplicates already-assigned referees by id.
+        def ref1 = Referee.builder()
+                .id(1L)
+                .averageGrade(8.1d)
+                .teamsRefereed(Map.of(team1, (short) 1, team2, (short) 1))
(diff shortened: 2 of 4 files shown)


src/main/java/com/jamex/refereestaffer/service/StafferService.java
```diff
@@ -67,7 +69,7 @@ private void assignRefereesToMatches(List<Referee> referees, List<Match> matches
                     .toList();
 
             var availableReferees = referees.stream()
-                    .filter(ref -> !ref.isBusy())
+                    .filter(ref -> !assignedRefereeIds.contains(ref.getId()))
```
UnLow1, line 72: Dedup is by id, so it assumes every referee here has a non-null, unique id. That holds in production — the list comes from `RefereeRepository` — but two *unsaved* entities would both map to a `null` key and the second would be wrongly excluded after the first assignment. If that assumption ever changes (e.g. staffing over detached/unsaved referees), this should switch to identity-based tracking.

Also worth noting: the vacation filter on the next line still relies on instance equality (`Referee` doesn't override `equals`). That's pre-existing behavior and works because vacations reference the same managed instances within one run — left untouched as out of scope for this ticket.

---

src/test/groovy/com/jamex/refereestaffer/service/StafferServiceSpec.groovy
```diff
@@ -82,12 +94,12 @@ class StafferServiceSpec extends Specification {
 
     def "should not assign referees to matches if referee has vacation"() {
         given:
-        def ref1 = [averageGrade: 8.6d] as Referee
-        def ref2 = [averageGrade: RefereeService.DEFAULT_GRADE, teamsRefereed: [:], numberOfMatchesInRound: 0] as Referee
-        def ref3 = [averageGrade: RefereeService.DEFAULT_GRADE, teamsRefereed: [:], numberOfMatchesInRound: 0] as Referee
-        def ref4 = [averageGrade: 8.6d] as Referee
-        def ref5 = [averageGrade: 8.6d] as Referee
-        def ref6 = [averageGrade: 8.6d] as Referee
+        def ref1 = Referee.builder().id(1L).averageGrade(8.6d).build()
```
UnLow1, line 97: These fixtures moved from map coercion (`[...] as Referee`) to the builder because the entity deliberately exposes no `setId` — the builder is the only way to give test referees ids, and ids are now required for the id-based dedup (with all-`null` ids, the staffer would exclude every remaining referee after the first assignment and throw `StafferException`). The two single-referee specs below keep map coercion since their referee is never assigned.

---

UnLow1 (commented):
Self-review notes on the two spots worth extra context.

---

src/main/java/com/jamex/refereestaffer/service/StafferService.java
UnLow1: Keeping the id-based `Set<Long>` as is — it's what RS-72 explicitly prescribes, and the non-null-id invariant is guaranteed by the only production call path (`RefereeService.getAvailableRefereesForQueue` → repository-loaded entities). Adding a defensive null check would guard against a state the application can't reach; if staffing ever operates on unsaved referees, that's a design change big enough to revisit this line anyway. No code change.

---

src/test/groovy/com/jamex/refereestaffer/service/StafferServiceSpec.groovy
UnLow1: This one is documentation for the reader rather than a change request — the fixture style split (builder with ids for multi-referee features, map coercion for the never-assigned ones) is deliberate and the suite is green, so no further change needed here.

---

PR #87, opened 2026-07-14
RS-73: Lock-aware staffing regenerate

UnLow1:
## Ticket

[RS-73: Lock-aware regenerate — staffReferees(queue, locks)](https://referee-staffer.youtrack.cloud/issue/RS-73)

The Staffer screen lets the user lock (pin) match–referee pairs, but the locks lived only in the UI: `POST /api/staffer/{queue}` knew nothing about them, so a regenerate could overwrite a pinned referee and the frontend had to restore locks after the fact (open TODO in `StafferComponent`). The ticket asks the endpoint to accept a list of pre-pinned `(matchId, refereeId)` pairs and treat them as taken — referee busy, match cast — while the algorithm staffs the rest.

## Plan

Analysis done before touching the code:

1. **Root problem is wider than the missing request body.** `staffReferees` is `@Transactional` and mutates managed entities, so assignments persist the moment a cast is generated (proven by `StafferIntegrationSpec`). The match query used `findAllByQueueAndRefereeIsNull`, so a second "Generate cast" found *nothing* to staff — regenerate returned an empty cast and could never reshuffle anything. Lock-aware regenerate therefore needs the backend to *reclaim* previously auto-assigned matches, not just accept locks.
2. **Redefine "matches to assign"** in `MatchService`: all matches of the queue except those that must keep their referee — central "S C" assignments and finished matches (both scores present). Unassigned matches stay included as before.
3. **Centralize the "S C" sentinel** check (`Referee.isCentralSentinel()`) instead of duplicating the magic name — the full flag-on-Match remodel suggested by the ticket is left as a follow-up.
4. **`StafferService.staffReferees(queue, locks)`**: validate locks → clear stale assignments → pin locked pairs → flush (the availability query is native SQL and only sees flushed state) → build the pool with the existing `getAvailableRefereesForQueue` (freed referees return, locked/kept ones drop out) → greedy-staff only the still-unassigned matches → return the whole assignable set sorted by difficulty.
5. **Controller**: optional `@RequestBody List<StaffingLockRequest>`; no body = plain regenerate, so existing clients keep working.
6. **Frontend**: send the lock map with the staffing request, clear locks on queue change (they reference matches of the previous queue), update the copy that said locks are "not yet preserved on regenerate".
7. **Tests**: Spock unit specs for pinning, re-staffing, and validation; integration specs proving lock survival and central-assignment immunity against real H2; controller spec for body parsing; Karma specs for the new frontend behavior.

Risks identified up front: semantic change of `getMatchesToAssignInQueue` (previously assigned, unfinished, non-central matches are now regenerated — that *is* the feature, and locking is the way to keep an assignment); locks deliberately bypass the vacation filter (an explicit user pin wins).

## Changes

- **`StafferController`** — `POST /api/staffer/{queue}` accepts an optional JSON array of `{matchId, refereeId}` locks (`StaffingLockRequest` record; `null` body → empty list).
- **`MatchService.getMatchesToAssignInQueue`** — now `findAllByQueue` + `isAssignable` filter: unassigned matches always; assigned ones unless central ("S C") or finished. `findAllByQueueAndRefereeIsNull` removed.
- **`Referee.isCentralSentinel()`** — single home for the "S C" magic-name check, reused in `RefereeService`.
- **`StafferService`** — lock validation (409 `StafferException` for non-assignable/duplicate locks, 404 `RefereeNotFoundException` for an unknown referee — all before any mutation), stale-assignment clearing, pinning, explicit `flush()` before the native availability query, auto-staffing restricted to still-unassigned matches. Returned cast includes the locked matches with their difficulty, sorted as before.
- **Transactionality fix** — the 1-arg `staffReferees` overload delegates via self-invocation, which bypasses the Spring proxy; it now carries its own `@Transactional`, otherwise assignments were never flushed (caught by the integration spec during development).
- **Frontend** — `StafferService.staffReferees(queue, locks)` posts the lock list; `StafferComponent.generate()` serializes the `locks` signal; `incQueue`/`decQueue` clear locks; comments and the "Locked assignments" KPI copy updated. New `StaffingLock` model interface.

Design decisions: validation errors reuse `StafferException` → existing 409 + ProblemDetail mapping (a lock conflict is a business-state conflict, not a malformed request); a failed staffing rolls the whole transaction back, so the previous cast survives; on error the toast comes from the existing global HTTP interceptor — no new UI needed.

## Testing

- Backend: `mvn -DskipFrontend=true test` — **141 tests green** (ran on OpenJDK 25). New: 5 unit features in `StafferServiceSpec` (pin + auto-staff split, stale-assignment clearing, non-assignable lock, duplicate lock pairs, unknown referee), 1 in `MatchServiceSpec` (assignability filter incl. central/finished), 1 in `StafferControllerSpec` (body → service parsing), 2 integration features in `StafferIntegrationSpec` (locked pair survives a regenerate and the rest is re-staffed from the remaining pool; central assignments untouched and excluded from the returned cast).
- Frontend: `npm ci`, `npm run lint` (clean), `npm test` headless — **139 tests green**. New: locks are sent on regenerate; queue change clears locks.

https://claude.ai/code/session_01HRnbstWjY9J8HvNUDw8AoR

---
_Generated by [Claude Code](https://claude.ai/code/session_01HRnbstWjY9J8HvNUDw8AoR)_

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/service/MatchService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/MatchService.java
@@ -3,6 +3,7 @@ package com.jamex.refereestaffer.service;
 import com.jamex.refereestaffer.model.dto.DifficultyBreakdownDto;
 import com.jamex.refereestaffer.model.entity.ConfigName;
 import com.jamex.refereestaffer.model.entity.Match;
+import com.jamex.refereestaffer.model.entity.Referee;
 import com.jamex.refereestaffer.model.entity.Team;
 import com.jamex.refereestaffer.model.exception.MatchNotFoundException;
 import com.jamex.refereestaffer.repository.ConfigurationRepository;
@@ -72,11 +73,25 @@ public class MatchService {
         matchRepository.delete(match);
     }
 
+    /**
+     * Matches the staffer is allowed to (re)assign in a queue, hardest first. Staffing
+     * persists assignments immediately, so a regenerate must be able to reclaim matches
+     * cast by a previous run — hence "assignable" covers previously assigned matches too,
+     * with two exceptions that keep their referee:
+     * <ul>
+     *   <li>central assignments (the "S C" sentinel — see {@link Referee#isCentralSentinel()}),</li>
+     *   <li>finished matches (both scores present) — history must not be rewritten.</li>
+     * </ul>
+     * Unassigned matches are always included, finished or not, matching the old
+     * referee-is-null behavior.
+     */
     public List<Match> getMatchesToAssignInQueue(Short queue) {
         var allFinishedMatches = matchRepository.findAllByHomeScoreNotNullAndAwayScoreNotNull();
         calculatePointsForTeams(allFinishedMatches);
 
-        var matchesToAssignInQueue = matchRepository.findAllByQueueAndRefereeIsNull(queue);
+        var matchesToAssignInQueue = matchRepository.findAllByQueue(queue).stream()
+                .filter(this::isAssignable)
+                .toList();
 
         // Config values and the team count are constant for the whole request — load them
         // once here instead of per match (used to be 4-5 findByName + count per iteration).
@@ -86,6 +101,16 @@ public class MatchService {
         return matchesToAssignInQueue.stream()
                 .sorted(Comparator.comparingDouble(Match::getHardnessLvl).reversed())
                 .toList();
+    }
+
+    private boolean isAssignable(Match match) {
+        if (match.getReferee() == null) {
+            return true;
+        }
+        if (match.getReferee().isCentralSentinel()) {
+            return false;
+        }
+        return match.getHomeScore() == null && match.getAwayScore() == null;
     }
 
     /**
--- a/src/main/java/com/jamex/refereestaffer/service/StafferService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/StafferService.java
@@ -7,8 +7,12 @@ import com.jamex.refereestaffer.model.entity.Match;
 import com.jamex.refereestaffer.model.entity.Referee;
 import com.jamex.refereestaffer.model.entity.Team;
 import com.jamex.refereestaffer.model.entity.Vacation;
+import com.jamex.refereestaffer.model.exception.RefereeNotFoundException;
 import com.jamex.refereestaffer.model.exception.StafferException;
+import com.jamex.refereestaffer.model.request.StaffingLockRequest;
 import com.jamex.refereestaffer.repository.ConfigurationRepository;
+import com.jamex.refereestaffer.repository.MatchRepository;
+import com.jamex.refereestaffer.repository.RefereeRepository;
 import com.jamex.refereestaffer.repository.VacationRepository;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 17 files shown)


src/main/java/com/jamex/refereestaffer/service/StafferService.java
```diff
@@ -18,43 +22,120 @@
...
+        var matchesById = matchesToStaff.stream()
+                .collect(Collectors.toMap(Match::getId, Function.identity()));
+        for (var lock : locks) {
+            var referee = refereeRepository.findById(lock.refereeId())
+                    .orElseThrow(() -> new RefereeNotFoundException(lock.refereeId()));
+            matchesById.get(lock.matchId()).setReferee(referee);
+        }
+    }
+
+    private void validateLocks(short queue, List<Match> matchesToStaff, List<StaffingLockRequest> locks) {
```
UnLow1, line 124: Known gap in this validation: it doesn't reject a lock whose referee already holds a *kept* (non-assignable) match in the same queue — e.g. a referee on a finished match, or the "S C" sentinel referee pinned via raw API. Such a lock would double-book the referee within the queue. Unreachable through the UI (the drawer's candidate list comes from `findAllWithNoMatchInQueue`, which excludes them), but reachable through the API.

---

src/main/java/com/jamex/refereestaffer/service/MatchService.java
```diff
@@ -88,6 +103,16 @@ public List<Match> getMatchesToAssignInQueue(Short queue) {
                 .toList();
     }
 
+    private boolean isAssignable(Match match) {
```
UnLow1, line 106: Deliberate semantic change worth calling out: a match that was assigned *manually* (via the match form) is now also reclaimed by "Generate cast" if it's unfinished and non-central — before this PR the staffer never touched any assigned match. This is what makes regenerate work at all (staffing persists immediately, so every previous auto-cast is "assigned"), and the backend can't tell manual assignments apart from auto ones. Locking the row is the supported way to keep an assignment through a regenerate. If manual assignments ever need first-class protection, that's the same modelling follow-up as the "S C" flag on `Match`.

---

src/main/java/com/jamex/refereestaffer/service/StafferService.java
```diff
@@ -18,43 +22,120 @@
...
+    }
+
+    /**
+     * (Re)generates the cast for a queue. Locked pairs pin a referee to a match up front:
+     * the match keeps that exact referee and the referee is unavailable for the rest of the
+     * cast. Every other assignable match (see {@link MatchService#getMatchesToAssignInQueue})
+     * is staffed from scratch, so a regenerate reshuffles previous auto-assignments instead
+     * of silently keeping them.
+     *
+     * <p>Locks deliberately bypass the vacation filter — a pinned pair is an explicit user
```
UnLow1, line 78: Design decision: a locked pair skips the vacation check on purpose. The user pinned it explicitly (usually after a manual swap in the drawer), and validating vacations here would make a previously valid lock fail the next regenerate, leaving the UI unable to restore its own state. The auto-staffing path still applies the vacation filter as before.

---

src/main/webapp/src/app/component/staffer/staffer.component.ts
```diff
@@ -101,10 +99,12 @@ export class StafferComponent {
 
   incQueue(): void {
     this.queue.update(q => q + 1);
+    this.clearLocks();
```
UnLow1, line 102: Trade-off: stepping to another queue and back drops the locks, even if the user never generated in between. The alternative (per-queue lock maps) felt like over-engineering for the current single-queue workflow, and stale locks sent for a different queue would now be rejected by backend validation (409), so clearing eagerly is the safer default.

---

UnLow1 (commented):
Self-review notes: flagged the lock-validation gap (double-booking via raw API), the regenerate semantic change for manually assigned matches, the deliberate vacation bypass for locks, and the lock-clearing trade-off on queue change.

---

src/main/java/com/jamex/refereestaffer/service/StafferService.java
UnLow1: Fixed in 942fabc: `validateLocks` now loads the queue's matches once (skipped for lock-less regenerates), derives the set of referees keeping a non-reassignable assignment, and rejects such a lock with 409 `LOCKED_REFEREE_UNAVAILABLE` before any state is mutated. Covered by a new unit feature ("should reject lock pinning a referee who keeps a non-reassignable match in the queue"). This also closes the raw-API path of pinning the "S C" sentinel referee whenever it holds a central match in the queue.

---

src/main/java/com/jamex/refereestaffer/service/MatchService.java
UnLow1: Keeping as-is. This reclaim behavior is the core of the ticket — without it a regenerate has nothing to reshuffle (assignments persist the moment a cast is generated). Distinguishing manual from auto assignments needs a flag on `Match`, which is the same modelling follow-up already noted for the "S C" sentinel; not expanding this PR's scope for it.

---

src/main/java/com/jamex/refereestaffer/service/StafferService.java
UnLow1: Keeping as-is. A lock is an explicit user override, and enforcing vacations on it would break the regenerate loop for locks created via drawer swaps (the drawer's candidate list is availability-based, not vacation-filtered). If vacation-aware locking is ever wanted, it belongs in the UI as a warning, not as a backend rejection.

---

src/main/webapp/src/app/component/staffer/staffer.component.ts
UnLow1: Keeping as-is. Eager clearing is the cheapest behavior that can never send a stale lock to the wrong queue, and losing a lock is recoverable in two clicks. Per-queue lock maps can be added later if the workflow ever becomes multi-queue.

---

PR #88, opened 2026-07-14
RS-74: Make Team.place a nullable Short instead of magic 0

UnLow1:
## Ticket

[RS-74: Team.place → Short (nullable) zamiast magic 0](https://referee-staffer.youtrack.cloud/issue/RS-74)

`Team.place` was a `@Transient` primitive `short` defaulting to `0` for teams that have no finished matches yet — a magic value. The defensive guard added to `MatchService.calculateHardnessLvlForEdgeMatch` (now `computeEdgeMatchParts`) special-cased `place == 0` so a fresh team's match would not accidentally be classified as a top-of-table fixture (`0 <= NUMBER_OF_EDGE_TEAMS`). That guard masked the modelling problem instead of fixing it. The ticket asks for the minimal fix: make `place` a nullable `Short` where `null` explicitly means "unranked".

## Plan

1. `Team.java`: change `@Transient short place` to `Short place` (null = unranked, never 0); update both constructors, getter/setter, and the nested `Builder`; document the semantics on the field.
2. `TeamConverter.convertFromDto`: build entities with `null` place instead of `(short) 0`.
3. `MatchService.computeEdgeMatchParts`: replace the `place == 0` guard with `place == null` and rewrite the comment to describe the new semantics.
4. `MatchServiceSpec`: represent unranked teams as `place: null` (previously `place: 0`) in both "unranked" data tables; add a converter spec assertion that a freshly converted entity is unranked.
5. Run `mvn -DskipFrontend=true test`.

Scope note: `TeamDto` never exposed `place`, so the REST contract and the Angular frontend are untouched. The ticket's further-reaching variant (moving `place`/`points` off the entity into a standings query) is explicitly left for RS-27.

## Changes

- `Team.java` — `place` is now `Short` with `null` meaning unranked; constructors, accessors, and `Builder` updated; javadoc added on the field. `points` intentionally stays a primitive `short` (0 points is a legitimate value, unlike place 0).
- `TeamConverter.java` — `convertFromDto` no longer fabricates `place = 0`; converted entities start unranked.
- `MatchService.java` — the edge-match guard checks `getPlace() == null`. `calculatePointsForTeams` still calls `setPlace((short) (i + 1))`, which now autoboxes; ranked places remain 1-based. The `place <= edge` / `place > teams - edge` comparisons unbox only after the null guard, so no NPE is possible.
- `MatchServiceSpec.groovy` — unranked rows in the two data tables use `null` instead of `0`, matching the new semantics; ranked rows unchanged.
- `TeamConverterSpec.groovy` — asserts `result.place == null` for DTO → entity conversion.

Also unchanged on purpose: `TeamService.getStandings` appends teams without matches to the standings list — those now carry `place = null` instead of `0`, but since `place` never leaves the backend, nothing downstream observes the difference.

## Testing

- `mvn -DskipFrontend=true test` — all 145 Spock tests pass.
- Coverage unchanged at 91.55% (Codecov); the changed lines are all covered, including the two parameterized "unranked" features (6 iterations) that pin the guard behaviour and the new converter assertion.
- No frontend changes, so the frontend workflow is not triggered (`paths` filter on `src/main/webapp/**`).

### Rebase note

Rebased onto master after #99 (DTOs → records) and the Java 26 bump landed. One conflict, in `TeamConverter.convertFromDto`: master switched the DTO reads to record accessors (`dto.id()`), this branch changed the trailing `place` argument to `null`. Resolved by keeping both — `new Team(dto.id(), dto.name(), dto.city(), null, (short) 0, null)`. Post-rebase suite is the 145 above (master's #99 added specs).

🤖 Generated with Claude Code

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/model/entity/Team.java
+++ b/src/main/java/com/jamex/refereestaffer/model/entity/Team.java
@@ -32,17 +32,22 @@ public class Team {
     @Transient
     private short points;
 
+    /**
+     * Standings position computed by {@code MatchService.calculatePointsForTeams} — only
+     * teams that appear in a finished match get ranked. {@code null} means unranked (no
+     * finished matches yet); never 0.
+     */
     @Transient
-    private short place;
+    private Short place;
 
     public Team() {
     }
 
-    public Team(Long id, String name, String city, short points, short place) {
+    public Team(Long id, String name, String city, short points, Short place) {
         this(id, name, city, null, points, place);
     }
 
-    public Team(Long id, String name, String city, String shortCode, short points, short place) {
+    public Team(Long id, String name, String city, String shortCode, short points, Short place) {
         this.id = id;
         this.name = name;
         this.city = city;
@@ -104,11 +109,11 @@ public class Team {
         return points;
     }
 
-    public short getPlace() {
+    public Short getPlace() {
         return place;
     }
 
-    public void setPlace(short place) {
+    public void setPlace(Short place) {
         this.place = place;
     }
 
@@ -135,7 +140,7 @@ public class Team {
         private String city;
         private String shortCode;
         private short points;
-        private short place;
+        private Short place;
 
         public Builder id(Long id) {
             this.id = id;
@@ -162,7 +167,7 @@ public class Team {
             return this;
         }
 
-        public Builder place(short place) {
+        public Builder place(Short place) {
             this.place = place;
             return this;
         }
--- a/src/main/java/com/jamex/refereestaffer/service/MatchService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/MatchService.java
@@ -138,11 +138,10 @@ public class MatchService {
 
     /** Returns [top, bottom] — at most one of them can be non-zero. */
     private double[] computeEdgeMatchParts(Team homeTeam, Team awayTeam, Map<ConfigName, Double> config, long numberOfTeams) {
-        // place is a @Transient field defaulting to 0 for any team that hasn't appeared in
-        // a finished match yet (calculatePointsForTeams only ranks teams from the finished set).
-        // Without this guard, place=0 satisfies `place <= numberOfTeamsOnEdge` (0 <= 3 by default),
-        // so a fresh team's match would be classified as a top-of-table fixture by accident.
-        if (homeTeam.getPlace() == 0 || awayTeam.getPlace() == 0) {
+        // place is null for any team that hasn't appeared in a finished match yet
+        // (calculatePointsForTeams only ranks teams from the finished set), so an unranked
+        // team can never be classified as a top- or bottom-of-table fixture.
+        if (homeTeam.getPlace() == null || awayTeam.getPlace() == null) {
             return new double[]{0.0, 0.0};
         }
         var numberOfTeamsOnEdge = config.get(ConfigName.NUMBER_OF_EDGE_TEAMS).longValue();
--- a/src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
+++ b/src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
@@ -113,9 +113,9 @@ class MatchServiceSpec extends Specification {
 
         where:
         homePoints | homePlace | awayPoints | awayPlace
-        0          | 0         | 0          | 0
-        0          | 0         | 30         | 1
-        30         | 1         | 0          | 0
+        0          | null      | 0          | null
+        0          | null      | 30         | 1
+        30         | 1         | 0          | null
     }
 
     def "should get matches to assign for given queue and set hardness level FULL"() {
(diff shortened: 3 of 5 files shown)


src/main/java/com/jamex/refereestaffer/model/entity/Team.java
```diff
@@ -32,17 +32,22 @@ public class Team {
...
     private short points;
 
+    /**
+     * Standings position computed by {@code MatchService.calculatePointsForTeams} — only
+     * teams that appear in a finished match get ranked. {@code null} means unranked (no
+     * finished matches yet); never 0.
+     */
     @Transient
-    private short place;
+    private Short place;
```
UnLow1, line 41: Deliberate asymmetry: `points` stays a primitive `short` while `place` becomes `Short`. For points, `0` is a legitimate value (a team with only losses genuinely has 0 points), so there is no magic value to eliminate. For place, positions are 1-based and `0` never meant anything real — hence only `place` gets the nullable treatment. RS-74 also floats the further-reaching option of moving both fields off the entity into a standings query; that is intentionally left for RS-27.

---

src/main/java/com/jamex/refereestaffer/service/MatchService.java
```diff
@@ -138,11 +138,10 @@ private boolean isDerby(Team homeTeam, Team awayTeam) {
...
     private double[] computeEdgeMatchParts(Team homeTeam, Team awayTeam, Map<ConfigName, Double> config, long numberOfTeams) {
-        // place is a @Transient field defaulting to 0 for any team that hasn't appeared in
-        // a finished match yet (calculatePointsForTeams only ranks teams from the finished set).
-        // Without this guard, place=0 satisfies `place <= numberOfTeamsOnEdge` (0 <= 3 by default),
-        // so a fresh team's match would be classified as a top-of-table fixture by accident.
-        if (homeTeam.getPlace() == 0 || awayTeam.getPlace() == 0) {
+        // place is null for any team that hasn't appeared in a finished match yet
+        // (calculatePointsForTeams only ranks teams from the finished set), so an unranked
+        // team can never be classified as a top- or bottom-of-table fixture.
+        if (homeTeam.getPlace() == null || awayTeam.getPlace() == null) {
```
UnLow1, line 144: The `getPlace() <= numberOfTeamsOnEdge` / `getPlace() > numberOfTeams - numberOfTeamsOnEdge` comparisons below unbox the `Short`, which would NPE for an unranked team — but they are only reachable after this null guard returns early, so both operands are guaranteed non-null there. The unranked Spock features (which deliberately omit the edge-config keys from the mocked config map) pin this ordering: if the guard ever moved below the config lookup or the comparisons, those tests would fail.

---

src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
```diff
@@ -113,9 +113,9 @@ class MatchServiceSpec extends Specification {
 
         where:
         homePoints | homePlace | awayPoints | awayPlace
-        0          | 0         | 0          | 0
-        0          | 0         | 30         | 1
-        30         | 1         | 0          | 0
+        0          | null      | 0          | null
```
UnLow1, line 116: These data tables mix `null` (unranked) and integer literals in the place columns. The `[place: awayPlace] as Team` map coercion goes through Groovy's `setProperty`, which passes `null` straight to `setPlace(Short)` and narrows `Integer` literals to `Short` via `castToType` — so no explicit `(short)` casts or `as Short` coercions are needed in the tables. Ranked rows (`place: 1`) keep working unchanged.

---

UnLow1 (commented):
Self-review notes on the discussable spots — context for the reader, no blocking issues found.

---

src/main/java/com/jamex/refereestaffer/model/entity/Team.java
UnLow1: Resolution: keeping as is. Widening this PR to also box `points` (or to move both fields off the entity) would go beyond RS-74's minimal-fix scope; the entity-level cleanup is tracked as the RS-27-adjacent variant mentioned in the ticket.

---

src/main/java/com/jamex/refereestaffer/service/MatchService.java
UnLow1: Resolution: no change needed. The guard-before-unboxing ordering is already enforced by the two parameterized "unranked" features, so the invariant is test-pinned rather than convention-only.

---

src/test/groovy/com/jamex/refereestaffer/service/MatchServiceSpec.groovy
UnLow1: Resolution: keeping the plain literals. Adding explicit `as Short` coercions to every row would be noise — Groovy's property-assignment coercion handles both `null` and `Integer` → `Short` reliably, and the suite passes on all 6 iterations of both tables.

---

PR #89, opened 2026-07-15
RS-75: Disable spring.jpa.open-in-view

UnLow1:
## Ticket

[RS-75](https://referee-staffer.youtrack.cloud/issue/RS-75) — `spring.jpa.open-in-view: false`

Open Session In View has been on by default since forever (Spring Boot logs a WARN about it on every startup). It holds the Hibernate session open for the whole HTTP request, which masks lazy-loading problems and lets hidden queries run during DTO serialization. The ticket asks to disable it and fix whatever surfaces in the service layer (fetch joins / DTO projections) rather than reverting the flag.

## Plan

1. **Analysis before touching anything:** map every association in the model and every place a lazy proxy could leak past a transaction boundary.
   - The only `LAZY` association is `Referee.matches` (`@OneToMany`) — and `getMatches()` is never called anywhere.
   - All `@ManyToOne` / `@OneToOne` associations are JPA-default `EAGER`, so they are loaded at repository-call time.
   - No `getReferenceById()` / `getOne()` usages (those return lazy proxies).
2. Set `spring.jpa.open-in-view: false` in the base `application.yml` (inherited by `dev` and `prod`).
3. Add an integration spec that drives full HTTP requests through the real stack (controller → converter/service → real repositories → H2 → JSON), since the existing controller specs are `@WebMvcTest` slices with mocked repositories and can never catch a session-scope regression.
4. Run the whole backend suite, then boot the app and exercise every read endpoint plus the staffer flow.
5. Fix anything that surfaces in the service layer.

## Changes

- **`application.yml`** — `spring.jpa.open-in-view: false`. Dev and prod inherit it; the startup WARN is gone.
- **`RefereeService.calculateStats`** — grouped the bulk match query by the `Referee` *entity*, relying on Hibernate handing back the same managed instances the controller passed in. That only held because OSIV gave the request a single session. Without it the two queries run in separate sessions, the instances differ, and — since `Referee` compares by identity — every lookup missed, silently resetting referee stats to defaults. Now grouped by `getReferee().getId()`.
- **`MatchService` — difficulty scoring** (second commit, from the review finding below). `computeDifficultyBreakdown` loads the match in one persistence context and then ranks the teams from a *second* repository call in another, so `computeBreakdown` read `points`/`place` off `Team` instances no ranking pass had touched: HTTP 200, no exception, every match scored with `pointsDiff = 0` and no table-zone bonus. `calculatePointsForTeams` now returns a `LeagueTable` keyed by team id and the scoring reads from that — the same id-over-identity pattern already applied to `calculateStats`.
- **`OpenInViewDisabledIntegrationSpec`** (new) — full-context `@SpringBootTest` + MockMvc spec asserting `/api/matches/{id}`, `/api/referees`, `/api/teams/standings`, `/api/vacations` and `/api/matches/{id}/difficulty`. The difficulty feature asserts concrete numbers, not a 200. `@Isolated` + `SAME_THREAD` because it shares H2 with `StafferIntegrationSpec` under Spock's parallel runner.
- **`MatchServiceSpec` / `RefereeServiceSpec`** — fixtures now carry ids, and the zone/derby permutations call `computeBreakdown` with an explicit `LeagueTable` instead of fabricating `points`/`place` directly on `Team` entities. Those fixtures described a state production cannot produce: the transient fields are only ever written by a ranking pass.

### Deliberately out of scope

The ranking rules are untouched — only teams with a finished match are ranked, ordering is by points alone, ties keep encounter order. Pointing the algorithm at `TeamService.getStandings` (which ranks *all* teams with a goal-difference tie-break chain) would move matches between table zones and change match hardness. Disabling OSIV must leave the numbers identical, so that unification — and deleting `Team.points` / `Team.place` — is tracked as [RS-99](https://referee-staffer.youtrack.cloud/issue/RS-99). For the same reason the entity mutation in `calculatePointsForTeams` stays for now.

## Testing

- CI is green on Java 26, including the Maven backend job that runs the Spock suite.
- Locally `mvn -DskipFrontend=true test` — 150/150 green.
- **The new difficulty feature was verified to actually fail without the fix**: reverting `computeBreakdown` to entity reads makes it return `{"total":100.0,"parts":{"base":100.0,...},"flags":{"isTop":false,"pointsDiff":0}}` and the spec fails on `pointsDiff == 3`. This is the check the first round of this PR was missing — it asserted status codes and relation fields, which a silently-wrong number sails straight through.
- Manual end-to-end run (default profile): seeded two teams, a referee, a finished 2:1 match and a grade through the API. `/api/matches/1/difficulty` returns `total 104.0, base 97.0, top 7.0, isTop true, pointsDiff 3` — matching the pre-OSIV-change reference values — and the staffer reports the same `hardnessLvl 104.0` for the equivalent fixture, so the two screens finally agree. No OSIV startup WARN, zero `LazyInitializationException` in the logs.

🤖 Generated with Claude Code

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/service/RefereeService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/RefereeService.java
@@ -54,13 +54,15 @@ public class RefereeService {
             return;
         }
         // One bulk query instead of a findAllByReferee per referee (N+1 — this runs on every
-        // referee list/profile GET, not just staffing). Grouping by the entity works because
-        // Hibernate returns the same managed Referee instances that went into the IN clause.
-        var matchesByReferee = matchRepository.findAllByRefereeIn(referees).stream()
-                .collect(Collectors.groupingBy(Match::getReferee));
+        // referee list/profile GET, not just staffing). Group by id, not by entity: with OSIV
+        // off, the referees passed in by a controller come from a different (already closed)
+        // session than this query, so Hibernate returns different instances — and Referee
+        // compares by identity, which would make every lookup below miss.
+        var matchesByRefereeId = matchRepository.findAllByRefereeIn(referees).stream()
+                .collect(Collectors.groupingBy(match -> match.getReferee().getId()));
 
         for (var referee : referees) {
-            var matchesForReferee = matchesByReferee.getOrDefault(referee, List.of());
+            var matchesForReferee = matchesByRefereeId.getOrDefault(referee.getId(), List.of());
 
             var averageGrade = countAverageGrade(matchesForReferee);
             var teamsRefereedMap = createTeamsRefereedMap(matchesForReferee);
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -2,6 +2,10 @@ spring:
   datasource:
     url: jdbc:h2:mem:testdb;MODE=MySQL;NON_KEYWORDS=VALUE
   jpa:
+    # Session-per-request (OSIV) masks lazy-loading problems and holds the Hibernate
+    # session open for the whole HTTP request. The only lazy association in the model
+    # (Referee.matches) is never traversed, so nothing relies on it — see RS-75.
+    open-in-view: false
     defer-datasource-initialization: true
     hibernate:
       ddl-auto: update
--- a/src/test/groovy/com/jamex/refereestaffer/integration/OpenInViewDisabledIntegrationSpec.groovy
+++ b/src/test/groovy/com/jamex/refereestaffer/integration/OpenInViewDisabledIntegrationSpec.groovy
@@ -1,0 +1,153 @@
+package com.jamex.refereestaffer.integration
+
+import com.jamex.refereestaffer.model.entity.Grade
+import com.jamex.refereestaffer.model.entity.Match
+import com.jamex.refereestaffer.model.entity.Referee
+import com.jamex.refereestaffer.model.entity.Team
+import com.jamex.refereestaffer.model.entity.Vacation
+import com.jamex.refereestaffer.repository.GradeRepository
+import com.jamex.refereestaffer.repository.MatchRepository
+import com.jamex.refereestaffer.repository.RefereeRepository
+import com.jamex.refereestaffer.repository.TeamRepository
+import com.jamex.refereestaffer.repository.VacationRepository
+import groovy.json.JsonSlurper
+import org.spockframework.runtime.model.parallel.ExecutionMode
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.boot.webmvc.test.autoconfigure.AutoConfigureMockMvc
+import org.springframework.core.env.Environment
+import org.springframework.test.web.servlet.MockMvc
+import spock.lang.Execution
+import spock.lang.Isolated
+import spock.lang.Specification
+
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get
+
+/**
+ * Regression net for disabling OSIV ({@code spring.jpa.open-in-view: false}, RS-75).
+ * The controller slices mock repositories away, so they can never catch a
+ * LazyInitializationException — this spec drives full HTTP requests (controller →
+ * converter/service → real repositories → H2 → JSON serialisation) with no
+ * session-per-request to lean on. If someone reintroduces a lazy traversal outside a
+ * transaction, these requests are where it blows up.
+ *
+ * {@code @Isolated} because it shares the H2 instance with other integration specs
+ * (parallel Spock run) and both sides wipe/seed domain tables in setup. {@code @Isolated}
+ * only fences off OTHER specs — features of this spec would still run concurrently and
+ * wipe each other's seed data, hence {@code SAME_THREAD} on top.
+ */
+@Isolated
+@Execution(ExecutionMode.SAME_THREAD)
+@AutoConfigureMockMvc
+@SpringBootTest
+class OpenInViewDisabledIntegrationSpec extends Specification {
+
+    @Autowired MockMvc mockMvc
+    @Autowired Environment environment
+    @Autowired TeamRepository teamRepository
+    @Autowired RefereeRepository refereeRepository
+    @Autowired MatchRepository matchRepository
+    @Autowired GradeRepository gradeRepository
+    @Autowired VacationRepository vacationRepository
+
+    def jsonSlurper = new JsonSlurper()
+
+    Long matchId
+    Long refereeId
+
+    def setup() {
+        wipeDomainData()
+
+        def home = teamRepository.save(new Team("Team1", "City1"))
+        def away = teamRepository.save(new Team("Team2", "City2"))
+        def referee = refereeRepository.save(new Referee("John", "Doe", "john@doe.com", 5))
+        short queue = 1
+        def match = matchRepository.save(new Match(queue, home, away,
+                LocalDateTime.of(2026, 3, 1, 12, 0), referee, (short) 2, (short) 1))
+        gradeRepository.save(Grade.builder().value(8.5d).match(match).build())
+        vacationRepository.save(new Vacation(null, referee,
+                LocalDate.of(2026, 7, 1), LocalDate.of(2026, 7, 14)))
+
+        matchId = match.id
+        refereeId = referee.id
+    }
+
+    def "should have OSIV disabled"() {
+        expect:
+        environment.getProperty("spring.jpa.open-in-view") == "false"
+    }
+
+    def "should serve a match with all relations resolved without OSIV"() {
+        when:
+        def response = mockMvc.perform(get("/api/matches/$matchId"))
+                .andReturn().response
+
+        then:
+        response.status == 200
+        // No with() here: bare property names inside with(parsedJson) resolve against the
+        // JSON delegate before spec fields, silently comparing the JSON to itself.
+        def match = jsonSlurper.parseText(response.contentAsString)
+        match.homeTeamId != null
+        match.awayTeamId != null
+        match.refereeId == refereeId
+        match.gradeId != null
+    }
+
+    def "should serve referees enriched with stats without OSIV"() {
+        when:
+        // enrichWithStats walks match → grade / home / away for every referee
+        def response = mockMvc.perform(get("/api/referees"))
+                .andReturn().response
+
+        then:
+        response.status == 200
+        def referees = jsonSlurper.parseText(response.contentAsString)
+        referees.size() == 1
+        referees[0].averageGrade == 8.5d
+        referees[0].potential != null
+    }
+
+    def "should serve standings computed from matches without OSIV"() {
+        when:
+        def response = mockMvc.perform(get("/api/teams/standings"))
+                .andReturn().response
+
+        then:
+        response.status == 200
+        def standings = jsonSlurper.parseText(response.contentAsString)
+        standings.size() == 2
+        standings[0].name == "Team1" // 2:1 home win → Team1 on top
+    }
+
+    def "should serve vacations with their referee resolved without OSIV"() {
+        when:
+        def response = mockMvc.perform(get("/api/vacations"))
+                .andReturn().response
+
+        then:
+        response.status == 200
+        def vacations = jsonSlurper.parseText(response.contentAsString)
+        vacations.size() == 1
+        vacations[0].refereeId == refereeId
+    }
+
+    def cleanup() {
+        // Leave only the data.sql config rows behind for whoever shares the H2 instance.
+        wipeDomainData()
+    }
+
+    private void wipeDomainData() {
+        // FK order: grade → match, match → team/referee, vacation → referee. Batch variants
+        // issue plain DELETEs without loading entities — deleteAll() would pull each Grade
+        // (and, through the EAGER one-to-one, its Match pointing back at the doomed Grade)
+        // into the session and fail the flush-time transient-reference check.
+        gradeRepository.deleteAllInBatch()
+        vacationRepository.deleteAllInBatch()
+        matchRepository.deleteAllInBatch()
+        refereeRepository.deleteAllInBatch()
+        teamRepository.deleteAllInBatch()
+    }
+}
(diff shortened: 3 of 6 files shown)


src/main/java/com/jamex/refereestaffer/service/RefereeService.java
```diff
@@ -54,13 +54,15 @@ public void calculateStats(List<Referee> referees) {
...
         // One bulk query instead of a findAllByReferee per referee (N+1 — this runs on every
-        // referee list/profile GET, not just staffing). Grouping by the entity works because
-        // Hibernate returns the same managed Referee instances that went into the IN clause.
-        var matchesByReferee = matchRepository.findAllByRefereeIn(referees).stream()
-                .collect(Collectors.groupingBy(Match::getReferee));
+        // referee list/profile GET, not just staffing). Group by id, not by entity: with OSIV
+        // off, the referees passed in by a controller come from a different (already closed)
+        // session than this query, so Hibernate returns different instances — and Referee
+        // compares by identity, which would make every lookup below miss.
+        var matchesByRefereeId = matchRepository.findAllByRefereeIn(referees).stream()
```
UnLow1, line 61: This is the one real OSIV dependency the ticket predicted. An alternative fix would be annotating `enrichWithStats`/`calculateStats` with `@Transactional`, but that wouldn't actually help here: the referees passed in by the controller were loaded *before* the transaction would start, so they'd still be detached instances distinct from the ones this query returns. Grouping by id is session-independent and keeps the method free of transaction demarcation it doesn't otherwise need.

Note the equivalent entity-keyed lookups in `StafferService` (`teamsRefereed.getOrDefault(homeTeam, …)`, `refereesWithVacations.contains(ref)`) are intentionally left untouched — `staffReferees` is `@Transactional`, so all its queries share one session and instance identity holds there.

---

src/main/resources/application.yml
```diff
@@ -2,6 +2,10 @@ spring:
   datasource:
     url: jdbc:h2:mem:testdb;MODE=MySQL;NON_KEYWORDS=VALUE
   jpa:
+    # Session-per-request (OSIV) masks lazy-loading problems and holds the Hibernate
+    # session open for the whole HTTP request. The only lazy association in the model
+    # (Referee.matches) is never traversed, so nothing relies on it — see RS-75.
+    open-in-view: false
```
UnLow1, line 8: Set in the base profile on purpose so `dev` and `prod` inherit it — tests, local runs, and the (future) deployment all behave the same. If a controller path ever genuinely needs a lazy traversal (e.g. someone starts using `Referee.matches`), the fix per the ticket is `@Transactional` / a fetch join / a DTO projection in the service layer — not flipping this back.

---

src/test/groovy/com/jamex/refereestaffer/integration/OpenInViewDisabledIntegrationSpec.groovy
```diff
@@ -0,0 +1,153 @@
...
+ * session-per-request to lean on. If someone reintroduces a lazy traversal outside a
+ * transaction, these requests are where it blows up.
+ *
+ * {@code @Isolated} because it shares the H2 instance with other integration specs
+ * (parallel Spock run) and both sides wipe/seed domain tables in setup. {@code @Isolated}
+ * only fences off OTHER specs — features of this spec would still run concurrently and
+ * wipe each other's seed data, hence {@code SAME_THREAD} on top.
+ */
+@Isolated
+@Execution(ExecutionMode.SAME_THREAD)
```
UnLow1, line 43: The pairing here is easy to get wrong (I got it wrong on the first run): `@Isolated` only takes a global lock at the *spec* level, so other specs can't overlap with this one — but the features inside still inherit Spock's `CONCURRENT` mode and were wiping each other's seed data mid-request. `SAME_THREAD` serializes the features. `StafferIntegrationSpec` gets away without either annotation only because it never ran concurrently with another H2-touching spec until now.

---

src/test/groovy/com/jamex/refereestaffer/integration/OpenInViewDisabledIntegrationSpec.groovy
```diff
@@ -0,0 +1,153 @@
...
+    def "should serve a match with all relations resolved without OSIV"() {
+        when:
+        def response = mockMvc.perform(get("/api/matches/$matchId"))
+                .andReturn().response
+
+        then:
+        response.status == 200
+        // No with() here: bare property names inside with(parsedJson) resolve against the
+        // JSON delegate before spec fields, silently comparing the JSON to itself.
+        def match = jsonSlurper.parseText(response.contentAsString)
```
UnLow1, line 92: Deliberately not using Spock's `with()` on parsed JSON anywhere in this spec: inside `with(map)` a bare name like `refereeId` resolves against the *delegate* (the JSON map) before the spec's own field, so `it.refereeId == refereeId` compares the JSON to itself and always passes. The first draft of this spec had exactly that tautology.

---

UnLow1 (commented):
Self-review notes — context for the discussion-worthy spots in the diff.

---

src/main/java/com/jamex/refereestaffer/service/RefereeService.java
UnLow1: Resolved as-is: keeping the id-based grouping. Also verified the null-key concern for `groupingBy` — `getReferee()` can't be null here because the query is `findAllByRefereeIn(referees)`, so every returned match has a referee from the IN list. The unit specs hit the null *id* case only because their mocked entities weren't persisted; giving them ids matches production reality.

---

src/main/resources/application.yml
UnLow1: No change needed — this is the deliberate decision: one switch in the base profile, verified against all three profiles' yml files (neither `dev` nor `prod` overrides `spring.jpa`beyond `ddl-auto`, so both inherit `open-in-view: false`).

---

src/test/groovy/com/jamex/refereestaffer/integration/OpenInViewDisabledIntegrationSpec.groovy
UnLow1: Checked whether `StafferIntegrationSpec` needs the same treatment now: it doesn't — this spec's `@Isolated` global lock means the two H2-touching specs can never overlap, and every other spec either mocks persistence (`@WebMvcTest` slices) or doesn't use Spring at all. If a third H2-touching integration spec ever appears, it should copy this annotation pair; noting that here rather than annotating `StafferIntegrationSpec` preemptively keeps this PR scoped to RS-75.

---

src/test/groovy/com/jamex/refereestaffer/integration/OpenInViewDisabledIntegrationSpec.groovy
UnLow1: Keeping the plain-assert style — it's slightly more verbose than `with()` but immune to the delegate-shadowing trap, and the inline comment warns the next person who's tempted to "clean it up". No code change.

---

UnLow1:
Finding from the RS-74 work: disabling OSIV also silently breaks `GET /api/matches/{id}/difficulty`. It is the same class of bug as the `RefereeService.calculateStats` regression this PR already fixes — cross-session instance identity — but in a different place, and `OpenInViewDisabledIntegrationSpec` does not cover it (it exercises `/api/matches/{id}`, `/api/referees`, `/api/teams/standings` and `/api/vacations`, but not the difficulty endpoint).

### Mechanism

`MatchService.computeDifficultyBreakdown` issues two separate repository calls:

1. `matchRepository.findById(matchId)` — loads the match with its teams;
2. `matchRepository.findAllByHomeScoreNotNullAndAwayScoreNotNull()` followed by `calculatePointsForTeams(...)`, which mutates `points`/`place` on the `Team` instances returned by **that second** query.

Nothing on this path is transactional (neither `MatchService` nor `MatchController` is annotated), so with `open-in-view: false` each repository call gets its own persistence context — `SimpleJpaRepository`'s methods are `@Transactional(readOnly = true)`. The `Team` instances from step 1 and step 2 are then different objects, and `computeBreakdown` reads `match.getHome()` / `match.getAway()`, i.e. the step-1 ones, which were never enriched: `points = 0`, `place = null`.

No exception is thrown. The endpoint returns HTTP 200 with wrong numbers: every match gets `pointsDiff = 0` (so an identical base difficulty) and never receives a top- or bottom-of-table bonus.

### Reproduction

Seeded two teams and one finished match 2:1 (Team1 = 3 pts, Team2 = 0 pts) and hit the endpoint through MockMvc on a full `@SpringBootTest` context:

| Branch | Response |
|---|---|
| `master` (OSIV on) | `{"total":104.0,"parts":{"base":97.0,"top":7.0,…},"flags":{"isTop":true,"pointsDiff":3}}` ✅ |
| `rs-75-disable-open-in-view` (OSIV off) | `{"total":100.0,"parts":{"base":100.0,"top":0.0,…},"flags":{"isTop":false,"pointsDiff":0}}` ❌ |

Same spec, same seed data, the only difference is the flag.

### Why the other paths survive

`staffReferees` is `@Transactional`, so the whole staffer path shares one session — as the PR description correctly notes. `getStandings` survives by accident rather than by design: it returns exactly the instances it mutated, so the session split never becomes observable. The difficulty endpoint is the one place where the match is loaded *before* and *separately from* the standings computation.

### Suggested fix

1. **Preferred** — pass places/points as a `Map<Long, …>` keyed by team id instead of reading them off the entity; `computeBreakdown` would look up `match.getHome().getId()`. That is precisely the pattern this PR already applied to `calculateStats` (group by id, not by entity), so it follows the precedent set here and does not depend on transaction boundaries.
2. **Minimum** — `@Transactional(readOnly = true)` on `computeDifficultyBreakdown`. One annotation, restores a single persistence context, but puts correctness back on the identity map, which is the very thing RS-75 is trying to lean on less.

Either way, `OpenInViewDisabledIntegrationSpec` deserves a feature for `/api/matches/{id}/difficulty` that asserts the actual numbers (with the 2:1 seed and the `data.sql` config: `pointsDiff == 3`, `parts.base == 97.0`). A status-200 assertion alone will not catch this, since the endpoint does not throw.

### Wider context

The root cause is that standings computation mutates managed entities instead of returning a read model: `addPoints` accumulates with no reset, and `Team` defines neither `equals` nor `hashCode`, so `.distinct()` and `HashMap<Team, …>` rely on the identity map too. Long term `points`/`place` should leave the entity in favour of a per-request projection — a natural fit for RS-27, which would also resolve RS-84. That would de-risk RS-75 properly, suggesting the order RS-27 → RS-75.

Housekeeping: this branch is still based on `e3e34e1`. Since then master has taken DTO records (#99), the Java 26 bump, and RS-74 (#88, `Team.place` is now a nullable `Short`), so a rebase is needed regardless. After RS-74 an unranked team carries `place = null` rather than `0`, but the outcome above is unchanged — the guard returns zero bonuses either way.

🤖 Generated with Claude Code

---

UnLow1:
Follow-up to my previous comment, after reviewing #83 (RS-27, standings on the backend) — it changes the recommended fix and the suggested ordering.

Above I proposed, as the preferred option, passing places into `computeDifficultyBreakdown` as a map keyed by `teamId`. It turns out **#83 already builds exactly that read model**: `TeamService.getStandings()` returns a `StandingsDto`, accumulating stats in a `HashMap<Long, TeamStats>` keyed by `teamId` and assigning `place = index + 1` in the DTO row, with no entity mutation. It is therefore inherently immune to the persistence-context split — precisely the mechanism that breaks the difficulty endpoint once OSIV is off.

So on ordering: **#83 first, then this PR.** The difficulty fix then stops being an ad-hoc map and becomes "consume the read model that already exists", which also removes the duplicated points logic and opens the door to deleting `Team.points` / `Team.place` from the entity altogether.

Important caveat: **#83 does not fix this bug by itself.** Verified in that branch's code — `MatchService.calculatePointsForTeams` is untouched, and both the staffer and difficulty paths still read the transient entity fields. After #83 merges there are two parallel league-table implementations: the read model behind `/api/teams/standings`, and the old entity mutation behind the algorithm. If this PR were to land before #83, the fix from my previous comment (map keyed by `teamId`, or `@Transactional(readOnly = true)`) applies unchanged.

Either way the test gap stands: `OpenInViewDisabledIntegrationSpec` needs a feature for `/api/matches/{id}/difficulty` asserting the actual numbers, since the endpoint does not throw and a status-200 check will not catch the regression.

🤖 Generated with Claude Code

---

UnLow1:
**Correction to my previous comment**, plus the resulting ordering.

I wrote above that once #83 lands, the difficulty fix "reduces to consuming the existing read model". **I'm withdrawing that recommendation.** I compared the two league-table implementations and they produce different places, so that swap would smuggle a behaviour change into the scoring algorithm under cover of a configuration fix:

| | old (`calculatePointsForTeams`) | new (`getStandings`, #83) |
|---|---|---|
| ranked teams | only those with a finished match | all teams from `findAll()` |
| sorting | points only | points → goal difference → goals for → name |
| ties | stream encounter order | explicit tie-break chain |
| team with no matches | unranked (`place = null` after RS-74) | real place at the bottom, zeroed stats |

A team that has not played would stop being unranked and could be classified as a bottom-of-table fixture, and ties would resolve differently — so match hardness would change.

**The fix in this PR should therefore be behaviour-preserving**: either a map keyed by `teamId` built from the *old* ranking logic (session-independent, consistent with the `calculateStats` precedent already set here), or minimally `@Transactional(readOnly = true)` on `computeDifficultyBreakdown`. Both leave today's numbers exactly as they are. Pointing the algorithm at the read model is now tracked separately as [RS-99](https://referee-staffer.youtrack.cloud/issue/RS-99), as an explicit, tested behaviour change in its own PR.

Rationale for keeping them apart: disabling OSIV is a config change that must leave the numbers untouched, whereas unifying the table changes the numbers. Mixed into one PR, any regression leaves you unable to tell whether the flag or the new zone classification caused it.

Suggested order for the cluster: **RS-27 (#83) → RS-84 → RS-75 (this PR) → RS-99 → RS-70**. #83 goes first because it is the enabler and its cost grows weekly (master is on Angular 22 + TS 6 while it targets 21); RS-84 rides along right after while that frontend is fresh; this PR then lands as a pure config change.

Still unchanged: `OpenInViewDisabledIntegrationSpec` needs a `/api/matches/{id}/difficulty` feature asserting concrete numbers — the endpoint does not throw, so a status-200 check will not catch the regression.

🤖 Generated with Claude Code

---

PR #90, opened 2026-07-15
RS-82: Focus trap in drawer and confirm dialog with focus return to trigger

UnLow1:
# Ticket

[RS-82 — Focus-trap w drawerze i confirm dialogu + powrót fokusu na trigger](https://referee-staffer.youtrack.cloud/issue/RS-82)

The post-redesign a11y baseline is good (`role="dialog"`/`aria-modal`, `alertdialog` on the confirm modal, Esc handling, focusable backdrops, `<label for>`, template-accessibility lint rules), but two details were missing: Tab could walk out of an open drawer/modal into the content underneath, and after closing an overlay the keyboard focus was left stranded instead of returning to the element that opened it. Both are required for the overlays to behave like real modal dialogs for keyboard and screen-reader users.

# Plan

1. Add a small shared focus-trap utility in `component/common/` — a `FocusTrapDirective` (`appFocusTrap`) shared by the only two overlay primitives in the app (`app-drawer`, `app-confirm-dialog`). Since both panels live inside `@if (open)` blocks, the directive activates on element creation and deactivates on destroy, so it needs no `open` input.
2. On activation: save `document.activeElement` (the trigger), move focus to the panel's first focusable element, and register in a module-level trap stack so stacked overlays (form drawer + its discard guard) don't fight — only the topmost trap handles Tab.
3. Trap Tab/Shift+Tab: wrap from last→first and first→last, pull focus back inside if it escaped, skip disabled controls.
4. On destroy: restore focus to the saved trigger, guarded by `isConnected` so that when a stacked guard closes together with its drawer, the drawer's restore (to the real trigger) wins.
5. Apply the directive to the drawer `<aside>` and the modal `<div>`; tests for the directive plus focus assertions in the drawer / confirm-dialog / form-drawer specs.
6. Verify with `npm ci`, `npm run lint`, `npm test` (headless).

# Changes

- **New `component/common/focus-trap/focus-trap.directive.ts`** — first directive in the app. Module-level `trapStack` (no injectable service needed; both primitives share one stack), document-level `keydown` listener per trap that only acts on Tab and only when the trap is topmost, `focusables()` based on a standard focusable selector that excludes disabled controls and `tabindex="-1"`.
- **`drawer.component.ts` / `confirm-dialog.component.ts`** — `appFocusTrap` applied to the panel elements. Design decision: the trap covers the *panel only*, so the focusable backdrop `<button>` drops out of the Tab cycle. That's intentional and standard for modal dialogs — Esc, the X button, and the footer buttons cover keyboard dismissal, while the backdrop button remains activatable by assistive tech and mouse. The stale comments documenting the old "Tab can land on the backdrop" behaviour were updated.
- Initial focus goes to the panel's first focusable element: the X button in drawers, the Cancel button in the confirm dialog (least destructive action first, per the dialog pattern).
- The `aria-live` item from the ticket was left out — the ticket marks it optional and ties it to toast changes, which are a separate concern.

# Testing

- New `focus-trap.directive.spec.ts` (9 specs): initial focus, Tab/Shift+Tab wrapping, disabled-control skipping, focus recapture after escaping, restore-to-trigger, stacked-trap behaviour (topmost wins, restore into the outer trap), and the simultaneous-close case where the inner trap's saved element is already detached.
- Extended `drawer.component.spec.ts` and `confirm-dialog.component.spec.ts` with open→focus-in / close→focus-back-to-trigger assertions, and `form-drawer.component.spec.ts` with a discard-guard stacking test (focus moves into the guard, returns to the drawer when the guard is dismissed).
- `npm run lint` clean, `npm test` headless: **149/149 SUCCESS** (12 new specs).

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01HVnxS6eeUigajuuWm8rpnm)_

### DIFF
--- a/src/main/webapp/src/app/component/common/drawer/drawer.component.ts
+++ b/src/main/webapp/src/app/component/common/drawer/drawer.component.ts
@@ -1,20 +1,23 @@
 import {Component, EventEmitter, HostListener, Input, Output} from '@angular/core';
 import {IconComponent} from '../icon/icon.component';
+import {FocusTrapDirective} from '../focus-trap/focus-trap.directive';
 
 /**
  * Right-side detail drawer with backdrop. Slide animation comes from styles.scss
  * (.drawer + @keyframes slide). Closes on backdrop click, on the X button, or on Esc.
+ * While open, appFocusTrap keeps Tab inside the panel and returns focus to the trigger
+ * on close.
  *
  * Slot content goes into .drawer__body. An optional footer can be slotted with
  * `<div drawerFoot>...</div>` and lands inside .panel__foot for consistent styling.
  */
 @Component({
   selector: 'app-drawer',
-  imports: [IconComponent],
+  imports: [IconComponent, FocusTrapDirective],
   template: `
     @if (open) {
       <button type="button" class="drawer-backdrop" (click)="closed.emit()" aria-label="Close drawer"></button>
-      <aside class="drawer" [class.drawer--wide]="wide" role="dialog" aria-modal="true" [attr.aria-label]="title">
+      <aside class="drawer" [class.drawer--wide]="wide" role="dialog" aria-modal="true" [attr.aria-label]="title" appFocusTrap>
         <header class="drawer__head">
           <div>
             <div class="drawer__title">{{ title }}</div>
@@ -37,9 +40,10 @@ import {IconComponent} from '../icon/icon.component';
     .drawer__title {
       font-weight: 600;
     }
-    /* The backdrop is a <button> for keyboard accessibility — it has to be focusable so
-       Tab can land on it and Enter/Space dismiss the drawer. Strip the default button
-       chrome (border, background) so it stays the dim overlay it visually has to be. */
+    /* The backdrop is a <button> so assistive tech can activate it as "Close drawer";
+       the focus trap keeps Tab inside the panel, so it is not part of the Tab cycle —
+       Esc and the X button cover keyboard dismissal. Strip the default button chrome
+       (border, background) so it stays the dim overlay it visually has to be. */
     button.drawer-backdrop {
       border: none;
       padding: 0;
--- a/src/main/webapp/src/app/component/common/focus-trap/focus-trap.directive.ts
+++ b/src/main/webapp/src/app/component/common/focus-trap/focus-trap.directive.ts
@@ -1,0 +1,84 @@
+import {AfterViewInit, Directive, ElementRef, OnDestroy, inject} from '@angular/core';
+
+/** Interactive elements a focus trap cycles through. Disabled controls are skipped. */
+const FOCUSABLE_SELECTOR = [
+  'a[href]',
+  'button:not([disabled])',
+  'input:not([disabled])',
+  'select:not([disabled])',
+  'textarea:not([disabled])',
+  '[tabindex]:not([tabindex="-1"])'
+].join(', ');
+
+/**
+ * Overlays can stack (form drawer + its discard guard), so every active trap registers
+ * here and only the topmost one handles Tab. Module-level on purpose: both overlay
+ * primitives share one stack without needing an injectable service.
+ */
+const trapStack: FocusTrapDirective[] = [];
+
+/**
+ * Traps Tab / Shift+Tab inside the host element and returns focus to the previously
+ * focused element (the trigger) when the host is destroyed. Designed for the two overlay
+ * primitives (app-drawer, app-confirm-dialog), whose panels live inside `@if (open)`
+ * blocks — creation activates the trap, removal deactivates it, so the directive needs
+ * no `open` input of its own.
+ *
+ * On activation focus moves to the first focusable element of the panel. The focusable
+ * backdrop button is a sibling of the panel, so it drops out of the Tab cycle while a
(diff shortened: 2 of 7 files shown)


src/main/webapp/src/app/component/common/focus-trap/focus-trap.directive.ts
```diff
@@ -0,0 +1,84 @@
...
+  'textarea:not([disabled])',
+  '[tabindex]:not([tabindex="-1"])'
+].join(', ');
+
+/**
+ * Overlays can stack (form drawer + its discard guard), so every active trap registers
+ * here and only the topmost one handles Tab. Module-level on purpose: both overlay
+ * primitives share one stack without needing an injectable service.
+ */
+const trapStack: FocusTrapDirective[] = [];
```
UnLow1, line 18: Deliberate choice: a module-level stack instead of an injectable `FocusTrapService`. With exactly two overlay primitives and at most two simultaneous traps (form drawer + discard guard), a service adds DI ceremony without buying anything — there is no per-injector state to isolate, and tests exercise the stack through the public behaviour anyway. If a third overlay type ever appears (per CLAUDE.md it shouldn't), this is the first thing to revisit.

---

src/main/webapp/src/app/component/common/focus-trap/focus-trap.directive.ts
```diff
@@ -0,0 +1,84 @@
...
+  private readonly onKeydown = (event: KeyboardEvent): void => this.trapTab(event);
+
+  ngAfterViewInit(): void {
+    this.previouslyFocused = document.activeElement instanceof HTMLElement ? document.activeElement : null;
+    trapStack.push(this);
+    document.addEventListener('keydown', this.onKeydown);
+
+    const target = this.focusables()[0] ?? this.container;
+    if (target === this.container) this.container.tabIndex = -1;
+    target.focus();
```
UnLow1, line 46: Initial focus lands on the panel's first focusable element — in drawers that is the X close button (the header precedes the body), in the confirm dialog the Cancel button (least destructive action, which is what the ARIA alertdialog pattern recommends). For form drawers one could argue the first *field* is friendlier, but that would need either an opt-in input on the directive or an `autofocus` convention in every form; first-focusable is the simpler and still spec-compliant default. Also note the selector doesn't filter by visibility (`display:none` etc.) — fine today because neither overlay hides focusable content, worth remembering if that changes.

---

src/main/webapp/src/app/component/common/drawer/drawer.component.ts
```diff
@@ -1,20 +1,23 @@
...
  */
 @Component({
   selector: 'app-drawer',
-  imports: [IconComponent],
+  imports: [IconComponent, FocusTrapDirective],
   template: `
     @if (open) {
       <button type="button" class="drawer-backdrop" (click)="closed.emit()" aria-label="Close drawer"></button>
-      <aside class="drawer" [class.drawer--wide]="wide" role="dialog" aria-modal="true" [attr.aria-label]="title">
+      <aside class="drawer" [class.drawer--wide]="wide" role="dialog" aria-modal="true" [attr.aria-label]="title" appFocusTrap>
```
UnLow1, line 20: Behaviour change worth flagging: the trap is on the `<aside>`, so the backdrop `<button>` (a sibling) is no longer reachable with Tab. Previously Tab could land on it and Enter dismissed the drawer; now keyboard dismissal is Esc / the X button, which is the standard modal-dialog pattern. The backdrop stays a `<button>` so it remains clickable and exposed to assistive tech — converting it to a plain `<div>` would be a bigger a11y regression than leaving a focusable-but-untabbable element around.

---

UnLow1 (commented):
Self-review notes on the discussion-worthy spots in this diff.

---

src/main/webapp/src/app/component/common/focus-trap/focus-trap.directive.ts
UnLow1: Resolved as a deliberate decision — keeping the module-level stack, no code change. The stack is private module state exercised only through the directive's observable behaviour (the stacking spec covers it), and a DI service would add indirection for a maximum depth of two.

---

src/main/webapp/src/app/component/common/focus-trap/focus-trap.directive.ts
UnLow1: Partially addressed in bf9c46f: `focusables()` now filters with `element.checkVisibility()`, so hidden controls can no longer become the initial-focus target or a wrap-around endpoint (covered by the `#invisible` button in the directive spec). The first-focusable initial-focus choice itself stays as is — an opt-in focus target can be added later if a form ever needs it.

---

src/main/webapp/src/app/component/common/drawer/drawer.component.ts
UnLow1: Keeping as is — this matches the WAI-ARIA dialog pattern (focus stays inside the dialog; Esc dismisses) and the backdrop keeps its click/AT affordance. The in-code comments on both backdrop styles were updated in this PR to document the new behaviour, so the old "Tab can land on it" rationale won't mislead anyone.

---

PR #92, opened 2026-07-16
RS-83: Add wildcard route and legacy /importer redirect

UnLow1:
## Ticket

[RS-83](https://referee-staffer.youtrack.cloud/issue/RS-83) — Wildcard route (** → dashboard) + redirect /importer → /import

Two routing gaps left over from the 2026-06 redesign review:

1. Navigating to an unknown URL threw the router's `NG04002` error to the console instead of landing the user anywhere useful.
2. The redesign renamed the `/importer` screen to `/import` without leaving a redirect behind, so pre-redesign bookmarks and old links died.

## Plan

1. Add `{path: 'importer', redirectTo: 'import'}` next to the `import` route in `app.routes.ts`, following the existing legacy deep-link convention (`/addReferee` etc. stay as-is — they intentionally open drawers).
2. Add `{path: '**', redirectTo: 'dashboard'}` as the **last** child of the shell route, so the redirect renders inside the shell and order-sensitive matching keeps all real routes ahead of it.
3. Cover both with a new `app.routes.spec.ts` that navigates the real route table (unknown URL → `/dashboard`, `/importer` → `/import`, known routes unaffected).

## Changes

- `src/main/webapp/src/app/app.routes.ts` — the two new route entries, each with a comment explaining why it exists (matching the commenting style already used in the file). Both redirects are relative to the `''` shell parent. The wildcard is a *child* of the shell rather than a sibling top-level route, so even a bad URL keeps the sidebar/topbar around it.
- `src/main/webapp/src/app/app.routes.spec.ts` — new spec exercising the real `routes` array through `RouterTestingHarness`. Since navigation actually renders the shell and the target screen, the services those components inject (`UiSettingsService`, `MatchService`, `RefereeService`, `TeamService`, `ImporterService`) are provided as spies/signal stubs, following the conventions from the RS-81 component-spec pass (PR #74).

Deliberate scope limits: `/importer/<anything-deeper>` falls through to the wildcard (the legacy screen never had child routes), and no other legacy paths were added — the ticket names only these two gaps.

## Testing

- `npm run lint` — clean.
- `npm test` (Karma, ChromeHeadlessNoSandbox) — 141/141 specs pass, including the 4 new routing specs.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01FuNK34a3XpNByuK1zBdMQV)_

### DIFF
--- a/src/main/webapp/src/app/app.routes.spec.ts
+++ b/src/main/webapp/src/app/app.routes.spec.ts
@@ -61,6 +61,13 @@ describe('app routes', () => {
     expect(router.url).toBe('/dashboard');
   });
 
+  it('sends deep paths under the legacy /importer to the dashboard, not /import', async () => {
+    // The old importer screen had no child routes, so only the exact path is preserved.
+    await harness.navigateByUrl('/importer/foo');
+
+    expect(router.url).toBe('/dashboard');
+  });
+
   it('redirects the legacy /importer path to /import', async () => {
     await harness.navigateByUrl('/importer');
 
--- a/src/main/webapp/src/app/app.routes.spec.ts
+++ b/src/main/webapp/src/app/app.routes.spec.ts
@@ -1,0 +1,75 @@
+import {TestBed} from '@angular/core/testing';
+import {signal} from '@angular/core';
+import {Router, provideRouter} from '@angular/router';
+import {RouterTestingHarness} from '@angular/router/testing';
+import {of} from 'rxjs';
+import {routes} from './app.routes';
+import {UiSettingsService} from './service/ui-settings.service';
+import {MatchService} from './service/match.service';
+import {RefereeService} from './service/referee.service';
+import {TeamService} from './service/team.service';
+import {ImporterService} from './service/importer.service';
+
+describe('app routes', () => {
+  let router: Router;
+  let harness: RouterTestingHarness;
+
+  beforeEach(async () => {
+    // Navigation renders the shell plus the target screen, so their services are
+    // stubbed the same way the component specs do it.
+    const matchService = jasmine.createSpyObj<MatchService>('MatchService', ['findAll']);
+    matchService.findAll.and.returnValue(of([]));
+    const refereeService = jasmine.createSpyObj<RefereeService>('RefereeService', ['findAll']);
+    refereeService.findAll.and.returnValue(of([]));
+    const teamService = jasmine.createSpyObj<TeamService>('TeamService', ['getStandings']);
+    teamService.getStandings.and.returnValue(of([]));
+    const importerService = jasmine.createSpyObj<ImporterService>('ImporterService', ['postFile', 'downloadExampleFile']);
+
+    const settings = {
+      dark: signal(false),
+      adminVisible: signal(false),
+      explainerVisible: signal(false),
+      toggleDark: jasmine.createSpy('toggleDark'),
+      toggleAdmin: jasmine.createSpy('toggleAdmin'),
+      toggleExplainer: jasmine.createSpy('toggleExplainer')
+    };
+
+    TestBed.configureTestingModule({
+      providers: [
+        provideRouter(routes),
+        {provide: UiSettingsService, useValue: settings},
+        {provide: MatchService, useValue: matchService},
+        {provide: RefereeService, useValue: refereeService},
+        {provide: TeamService, useValue: teamService},
+        {provide: ImporterService, useValue: importerService}
+      ]
+    });
+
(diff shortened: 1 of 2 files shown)


src/main/webapp/src/app/app.routes.ts
```diff
@@ -48,6 +50,8 @@ export const routes: Routes = [
       {path: 'vacations', component: VacationListComponent},
       {path: 'addVacation', component: VacationListComponent},
       {path: 'addVacation/:id', component: VacationListComponent},
+      // Must stay last — catches any unknown URL instead of throwing NG04002 to the console.
+      {path: '**', redirectTo: 'dashboard'},
```
UnLow1, line 54: Design note: this is a *silent* redirect — a typo'd URL just lands on the dashboard with no explanation. The alternative would be a dedicated 404 screen, but the ticket explicitly asked for `** → dashboard` and the app has no concept of a not-found page today. If that ever changes, this is the line to swap to a `NotFoundComponent`.

Also worth calling out: the wildcard lives *inside* the shell's children (not as a top-level sibling), so the redirect keeps the sidebar/topbar mounted. Since the parent path `''` prefix-matches every URL, a top-level wildcard would never be needed anyway.

---

src/main/webapp/src/app/app.routes.ts
```diff
@@ -39,6 +39,8 @@ export const routes: Routes = [
       {path: 'addReferee/:id', component: RefereeListComponent},
       {path: 'grades', component: GradeListComponent},
       {path: 'import', component: ImporterComponent},
+      // Legacy path — the redesign renamed /importer to /import; keep old bookmarks alive.
+      {path: 'importer', redirectTo: 'import'},
```
UnLow1, line 43: This only matches `/importer` exactly — anything deeper (e.g. `/importer/foo`) falls through to the wildcard and ends up on the dashboard. That's intentional: the pre-redesign importer screen never had child routes or an `:id` variant, so there is nothing deeper to preserve.

---

src/main/webapp/src/app/app.routes.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  beforeEach(async () => {
+    // Navigation renders the shell plus the target screen, so their services are
+    // stubbed the same way the component specs do it.
+    const matchService = jasmine.createSpyObj<MatchService>('MatchService', ['findAll']);
+    matchService.findAll.and.returnValue(of([]));
+    const refereeService = jasmine.createSpyObj<RefereeService>('RefereeService', ['findAll']);
+    refereeService.findAll.and.returnValue(of([]));
+    const teamService = jasmine.createSpyObj<TeamService>('TeamService', ['getStandings']);
+    teamService.getStandings.and.returnValue(of([]));
+    const importerService = jasmine.createSpyObj<ImporterService>('ImporterService', ['postFile', 'downloadExampleFile']);
```
UnLow1, line 26: The spy list is driven by which components actually render during these navigations: shell (`UiSettingsService`), dashboard (`MatchService`/`RefereeService`/`TeamService` via the `forkJoin` in `ngOnInit`) and importer (`ImporterService`, never called on init but still injected). If a future route test navigates to another screen, its services need to be added here — the failure mode is a `NullInjectorError` naming the missing service, so it's self-diagnosing.

---

UnLow1 (commented):
Self-review notes on the debatable spots — see inline comments.

---

src/main/webapp/src/app/app.routes.ts
UnLow1: Since this fall-through is a deliberate decision rather than an accident, it's now pinned by a dedicated spec in cb23396 (`/importer/foo` → `/dashboard`), so a future change to this behavior has to be conscious. Route entry itself stays as-is.

---

src/main/webapp/src/app/app.routes.ts
UnLow1: Keeping the silent redirect as-is: it's exactly what RS-83 asks for, and a 404 screen would be new scope (new component, nav decisions) that should get its own ticket if ever wanted. The comment on the route marks the spot to swap in a `NotFoundComponent` later.

---

src/main/webapp/src/app/app.routes.spec.ts
UnLow1: No change needed — this mirrors the RS-81 convention of stubbing exactly what renders, and the in-file comment above the spies already explains the rule. A missing spy fails loudly with a `NullInjectorError` naming the service, so there's no silent-failure risk to guard against.

---

PR #93, opened 2026-07-16
RS-84: Drive edge-team zones from backend configuration

UnLow1:
## Ticket

[RS-84 — NUMBER_OF_EDGE_TEAMS z API zamiast hardcode w 4 komponentach](https://referee-staffer.youtrack.cloud/issue/RS-84)

The frontend had `NUMBER_OF_EDGE_TEAMS = 3` hardcoded as a module-level const in four components (dashboard, staffer, match-detail, standings), while the backend keeps the value configurable in the `config` table and exposes it via `GET /api/configuration`. Any change of the config on the Configuration screen would silently desynchronize the locally derived zones/flags from the server-side computation. After RS-27 (#83) the `place` values come from the backend (`StandingsDto` rows), but the zone *thresholds* were still derived from the duplicated local constant — this PR removes that last local derivation.

## Plan

1. **`ConfigurationService`** — add an `edgeTeams` signal (fallback `3`, matching the `data.sql` seed) plus an idempotent `ensureEdgeTeamsLoaded()` that fetches `/api/configuration` once ("jeden strzał, w serwisie z sygnałem" per the ticket). Additionally tap `findAll()`/`update()` responses so the signal stays in sync whenever the Configuration screen loads or saves.
2. **`dashboard`** — drop the local const; `zoneFor()` reads the signal; trigger the one-shot load in `ngOnInit`.
3. **`standings`** — drop the const; `filterOptions` becomes a `computed` so the "Top N / Bottom N" segment labels follow the config; zone math on the backend-provided `place` reads the signal; load in `ngOnInit`.
4. **`match-detail`** — the local fallbacks in `isTopMatch`/`isRelegationMatch` (used only until the backend breakdown lands) read the signal; load in `ngOnInit`.
5. **`staffer`** — `flags()` reads the signal; load in the constructor (the screen fetches its data lazily on Generate, but the flag derivation is local).
6. Make the hardcoded "top-3" chip/row labels in the staffer and match-detail templates follow the configured value.
7. Tests: extend the service spec and the component specs with a `ConfigurationService` stub and tests proving zones/labels follow the configured value.

## Changes

- `ConfigurationService`: new exported `DEFAULT_NUMBER_OF_EDGE_TEAMS = 3` fallback, readonly `edgeTeams` signal, `ensureEdgeTeamsLoaded()` (one-shot; a failed fetch keeps the fallback and retries on the next call; any successful `findAll()`/`update()` — e.g. from the Configuration screen — counts as loaded, so no duplicate request), and a private `syncEdgeTeams()` applied via `tap` on both `findAll()` and `update()`. Values are `Math.round`-ed (backend stores doubles) and non-positive/missing values are ignored so a bad config can't break the UI.
- All four components inject the service, expose `readonly edgeTeams` for their templates, and derive zones/flags from it instead of a module const. Load is triggered from `ngOnInit` (dashboard, standings, match-detail) or the constructor (staffer).
- Templates: standings binds `filterOptions()` (now a computed); "top-3" chip and "Top-3 match" breakdown-row labels in staffer/match-detail render `top-{{ edgeTeams() }}`.
- Design decisions: kept the signal in the existing `ConfigurationService` rather than adding a new service; lazy idempotent load from consumers instead of an eager app-bootstrap fetch; non-blocking fetch — components render immediately with the fallback and re-derive when the response lands (signals propagate automatically).
- **Rebased onto master after RS-27 (#83), RS-79 (Angular 22), RS-85, RS-89 and RS-80 landed.** Zone math in standings/staffer/match-detail now uses the backend-provided `place` from `StandingsDto` rows (RS-27) with only the threshold coming from the config signal. The RS-89 specs for dashboard/match-detail and the RS-80 `configuration.service.spec.ts` gained the `ConfigurationService` stub / `edgeTeams` test block instead of introducing parallel spec files.
- Scope note: the RS-84 ticket also mentions the alternative of returning ready zone flags in `StandingsDto` rows. The config-signal approach was kept because the edge size is needed beyond the standings rows anyway — for the staffer/match-detail local flag fallbacks and for the "Top N / Bottom N / top-N" labels — so a single shared signal covers all four screens without a backend change.

## Testing

- `configuration.service.spec.ts` (on top of the RS-80 spec): fallback before fetch, one-shot + idempotence, skip when already loaded elsewhere, retry after error, ignoring missing/non-positive values, sync on `update()`.
- Component specs (`standings`, `staffer`, `dashboard`, `match-detail`): stubbed `ConfigurationService` (writable signal, same pattern as the `UiSettingsService` stub) + new tests that zones, filter labels and flag fallbacks follow a changed edge size (set to 2).
- `npm run lint` — clean; `npm test` (ChromeHeadlessNoSandbox, Node 26.5.1) — 258/258 green; `npm run build` (Angular 22 esbuild, production) — passes.

🤖 Generated with Claude Code

### DIFF
--- a/src/main/webapp/src/app/component/staffer/staffer.component.ts
+++ b/src/main/webapp/src/app/component/staffer/staffer.component.ts
@@ -4,6 +4,7 @@ import {StafferService} from '../../service/staffer.service';
 import {TeamService} from '../../service/team.service';
 import {RefereeService} from '../../service/referee.service';
 import {MatchService} from '../../service/match.service';
+import {ConfigurationService} from '../../service/configuration.service';
 import {UiSettingsService} from '../../service/ui-settings.service';
 import {Match} from '../../model/match';
 import {Team} from '../../model/team';
@@ -29,8 +30,6 @@ interface Candidate {
   isUsedElsewhere: boolean;
 }
 
-const NUMBER_OF_EDGE_TEAMS = 3;
-
 /**
  * Staffer — the auto-assignment workspace. Pick a queue, generate the cast, lock or
  * swap individual rows, then save.
@@ -53,8 +52,12 @@ export class StafferComponent {
   private readonly teamService = inject(TeamService);
   private readonly refereeService = inject(RefereeService);
   private readonly matchService = inject(MatchService);
+  private readonly configurationService = inject(ConfigurationService);
   /** Gates the "How the staffer scores assignments" panel (toggled in the sidebar's Admin section). */
   readonly settings = inject(UiSettingsService);
+
+  /** Edge-zone size (NUMBER_OF_EDGE_TEAMS) from the backend configuration. */
+  readonly edgeTeams = this.configurationService.edgeTeams;
 
   readonly queue = signal(1);
   readonly matches = signal<Match[] | null>(null);
@@ -77,6 +80,10 @@ export class StafferComponent {
   readonly drawerBreakdown = signal<DifficultyBreakdown | null>(null);
   readonly savedAt = signal<Date | null>(null);
   readonly loading = signal(false);
+
+  constructor() {
+    this.configurationService.ensureEdgeTeamsLoaded();
+  }
 
   // ——— Derived state ———
 
@@ -212,10 +219,11 @@ export class StafferComponent {
     const awayPlace = this.placeById().get(match.awayTeamId);
     const total = this.totalTeams();
 
+    const edge = this.edgeTeams();
     const sameCity = !!(home?.city && away?.city && home.city === away.city);
-    const isTop = !!(homePlace && awayPlace && homePlace <= NUMBER_OF_EDGE_TEAMS && awayPlace <= NUMBER_OF_EDGE_TEAMS);
+    const isTop = !!(homePlace && awayPlace && homePlace <= edge && awayPlace <= edge);
     const isBot = !!(homePlace && awayPlace && total > 0
-      && homePlace > total - NUMBER_OF_EDGE_TEAMS && awayPlace > total - NUMBER_OF_EDGE_TEAMS);
+      && homePlace > total - edge && awayPlace > total - edge);
     return {sameCity, isTop, isBot};
   }
 
--- a/src/main/webapp/src/app/component/standings/standings.component.ts
+++ b/src/main/webapp/src/app/component/standings/standings.component.ts
@@ -1,10 +1,9 @@
 import {Component, OnInit, computed, inject, signal, ChangeDetectionStrategy} from '@angular/core';
 import {Standing} from '../../model/standing';
 import {TeamService} from '../../service/team.service';
+import {ConfigurationService} from '../../service/configuration.service';
 import {TeamPillComponent} from '../common/team-pill/team-pill.component';
 import {SegComponent, SegOption} from '../common/seg/seg.component';
-
-const NUMBER_OF_EDGE_TEAMS = 3;
 
 export type StandingsFilter = 'all' | 'top' | 'bottom';
 
@@ -19,8 +18,8 @@ interface StandingRow {
  *
  * The whole table (order, points, place, P/W/D/L/GF/GA and the "after queue N"
  * subtitle) comes from /api/teams/standings — nothing is derived client-side anymore.
- * Top 3 = accent zone, bottom 3 = danger, matching the dashboard widget's
- * NUMBER_OF_EDGE_TEAMS.
+ * Zone sizes follow NUMBER_OF_EDGE_TEAMS from the backend configuration (top edge =
+ * accent, bottom edge = danger), matching the dashboard widget.
  */
 @Component({
   selector: 'app-standings',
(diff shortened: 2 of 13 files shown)


src/main/webapp/src/app/service/configuration.service.ts
```diff
@@ -18,11 +21,42 @@ export class ConfigurationService {
...
+   * every screen that derives edge zones calls this on init and only the first call hits
+   * the backend. On error the fallback stays and the next call retries.
+   */
+  public ensureEdgeTeamsLoaded(): void {
+    if (this.edgeTeamsLoaded) return;
+    this.edgeTeamsLoaded = true;
+    this.findAll().subscribe({
+      error: () => this.edgeTeamsLoaded = false
+    });
+  }
```
UnLow1, line 44: Deliberate error-handling choice: the guard flag is set optimistically before the request and reset in the error callback, so a failed fetch keeps the fallback (3) for the current screen but the next screen navigation retries. The alternative — a shared `ReplaySubject`/`shareReplay` pipeline — would cache a failure forever or need extra retry plumbing; a boolean is enough for a value this small. Note the HTTP error still flows through the global toast interceptor, so a config outage is visible, not swallowed.

---

src/main/webapp/src/app/service/configuration.service.ts
```diff
@@ -18,11 +21,42 @@ export class ConfigurationService {
...
     return this.http.get<Config[]>(this.configurationUrl)
+      .pipe(tap(configs => this.syncEdgeTeams(configs)))
+  }
+
+  private syncEdgeTeams(configs: Config[]): void {
+    const value = configs.find(c => c.name === 'NUMBER_OF_EDGE_TEAMS')?.value;
+    if (value != null && value > 0) {
+      this.edgeTeamsSignal.set(Math.round(value));
+    }
   }
```
UnLow1, line 64: `Math.round` because the backend stores every config as a double (`3.0` in `data.sql`) while zone math needs an integer; the `> 0` guard means a missing key or a nonsensical value (0, negative) leaves the last known good value in place instead of wiping the zones. This also runs on `update()`, so saving on the Configuration screen refreshes the zones everywhere without a reload — that's why the sync lives in the service and not in the consuming components.

---

src/main/webapp/src/app/component/staffer/staffer.component.ts
```diff
@@ -77,6 +80,10 @@ export class StafferComponent {
   readonly savedAt = signal<Date | null>(null);
   readonly loading = signal(false);
 
+  constructor() {
+    this.configurationService.ensureEdgeTeamsLoaded();
+  }
```
UnLow1, line 86: Staffer is the one screen without `ngOnInit` (it loads data lazily on Generate), so the one-shot config load goes in the constructor. It is intentionally NOT part of the `generate()` forkJoin: the fetch is non-blocking by design — `flags()` renders with the fallback and re-derives automatically when the signal updates, and in practice the config has already been fetched by the time a cast is generated. Note the chips here are a local derivation only; the authoritative difficulty increments in the drawer come from the backend breakdown.

---

src/main/webapp/src/app/component/standings/standings.component.ts
```diff
@@ -41,16 +41,20 @@ interface StandingRow {
...
 
-  readonly filterOptions: SegOption<StandingsFilter>[] = [
+  readonly filterOptions = computed<SegOption<StandingsFilter>[]>(() => [
     {value: 'all', label: 'All'},
-    {value: 'top', label: `Top ${NUMBER_OF_EDGE_TEAMS * 2}`},
-    {value: 'bottom', label: `Bottom ${NUMBER_OF_EDGE_TEAMS}`}
-  ];
+    {value: 'top', label: `Top ${this.edgeTeams() * 2}`},
+    {value: 'bottom', label: `Bottom ${this.edgeTeams()}`}
+  ]);
```
UnLow1, line 46: `filterOptions` had to become a `computed` (template now binds `filterOptions()`) so the "Top N / Bottom N" labels re-render when the config lands — with a plain array the zone rows would follow the config but the segment labels would stay frozen at the fallback. The `top` filter keeps the pre-existing `edge * 2` convention ("Top 6" for edge 3), matching the old hardcoded behavior.

---

UnLow1 (commented):
Self-review notes — context for the discussion-worthy spots in this diff.

---

src/main/webapp/src/app/service/configuration.service.ts
UnLow1: Follow-up applied in 6de14ff: reviewing this flow surfaced one gap — the Configuration screen calls `findAll()` directly, which didn't set the guard flag, so visiting Configuration first still let the next screen fire a duplicate `/api/configuration` request. `syncEdgeTeams()` now marks the config as loaded on any successful response, with a spec covering it. The optimistic-set/reset-on-error retry semantics described here stay as-is.

---

src/main/webapp/src/app/service/configuration.service.ts
UnLow1: No code change needed beyond 6de14ff (which additionally marks the config as loaded here). The rounding + positive-value guard is a conscious defensive choice: the Configuration screen allows saving arbitrary numbers and a 0/negative edge size would silently blank every zone in the UI, so keeping the last known good value is the safer failure mode.

---

src/main/webapp/src/app/component/staffer/staffer.component.ts
UnLow1: Keeping as-is — constructor load is the right trade-off here. Folding the config fetch into the `generate()` forkJoin would couple cast generation to an unrelated request (and delay it), while the constructor variant costs nothing and the signal re-derives the flags automatically in the unlikely window where the response lands after a cast.

---

src/main/webapp/src/app/component/standings/standings.component.ts
UnLow1: Keeping as-is. The `edge * 2` "Top N" convention is preserved deliberately — changing what the top filter shows is a product decision outside RS-84's scope (the ticket is about eliminating the hardcode, not redefining the zones). The standings spec pins both the labels and the row filtering for a non-default edge size.

---

UnLow1:
Rebased onto current master (`42c2587`, post RS-27/RS-79/RS-85/RS-89/RS-80) — same two commits, adapted to the new base:

- zone math now runs on the backend-provided `place` from `StandingsDto` rows (RS-27) with only the edge-size threshold coming from the config signal; the standings component keeps master's single-`getStandings()` shape (no `forkJoin`/`MatchService` resurrected);
- the `edgeTeams` test block merged into the RS-80 `configuration.service.spec.ts` instead of a parallel spec file, and the RS-89 dashboard/match-detail specs got the `ConfigurationService` stub plus an edge-size test each;
- templates re-applied on top of the RS-85 responsive markup.

Full verification on the Angular 22 toolchain: lint clean, 258/258 unit tests green, production build passes. (The inline replies below reference pre-rebase commit `6de14ff`; that change now lives in `711fb38`.)

🤖 Generated with Claude Code

---

PR #94, opened 2026-07-17
RS-85: Responsive pass — make the UI usable below 768px

UnLow1:
## Ticket

[RS-85: Responsive pass — użyteczność poniżej 768px](https://referee-staffer.youtrack.cloud/issue/RS-85)

The 2026-06 redesign shipped a desktop-only UI: the shell uses a rigid grid with a fixed 240px sidebar, data tables carry fixed column widths that force page-level horizontal scrolling on narrow viewports (standings is the worst offender with 12 columns), drawers are fixed at 440/560px (wider than most phones), and multi-column grids (dashboard split, KPI strips, staffer explainer, form rows) never collapse. There was not a single `@media` query in the codebase. The task: make the app usable below 768px.

## Plan

1. **Shell → off-canvas nav.** Add a single 768px breakpoint. Below it, drop the fixed sidebar column from the shell grid and turn `.sidenav` into a fixed off-canvas panel under the 56px topbar, toggled by a new hamburger button (`navOpen` signal in `ShellComponent`). Close on nav-item click, backdrop click, and Escape. Hide the static season/round topbar labels on narrow screens.
2. **Tables → scroll containers.** Wrap every `<table class="tbl">` (14 tables across 10 components) in a `.tbl-scroll` container with `overflow-x: auto`, so wide tables scroll inside their panel instead of stretching the page.
3. **Drawers → full width.** Detail and form drawers (`.drawer`, `.drawer--wide`) span the full viewport below the breakpoint.
4. **Grids → stack.** Collapse the global form grids (`.form-grid`, `.form-row.cols-2/3`, `.vs-split`) and the screen-specific grids: dashboard split + KPI strip, staffer explainer + KPI strip, referee-profile KPI strip, importer controls/result, configuration grid, match-detail hero + body. Let `.page-head` action rows wrap under the title.
5. **Tests.** Extend the `ShellComponent` spec with the mobile-nav interactions; run lint, unit tests, and a production build; verify visually at 375px with a headless browser.

## Changes

- `styles.scss` — new `.tbl-scroll` utility plus a single `@media (max-width: 767px)` block: shell grid collapse, off-canvas sidenav + backdrop (z-index 44/45, deliberately below `.drawer` at 50 and `.modal-backdrop` at 60 so overlays stack above the nav), full-width drawers, stacked form grids, wrapping page heads.
- `ShellComponent` — `navOpen` signal, `toggleNav()`/`closeNav()`, Escape `HostListener`; template gains the hamburger button (`aria-expanded`/`aria-controls`), the backdrop (a `<button>`, same keyboard-accessible pattern as the drawer backdrop), and `(click)="closeNav()"` on nav links; component SCSS hides the hamburger on desktop and the decorative topbar labels on mobile.
- `IconComponent` — new `menu` glyph (same three-line path as `list`; separate name so the toggle reads as "menu" at the call site).
- 10 component templates — tables wrapped in `.tbl-scroll` (markup re-indented, no other changes).
- 6 component stylesheets — screen-specific grid collapses at the same 767px breakpoint (KPI strips fold 4→2 and 5→2, splits stack, match-detail hero centers).
- `CLAUDE.md` — documented the breakpoint convention and the "wrap new tables in `.tbl-scroll`" rule.

Design decisions: a single breakpoint (per the ticket) instead of a fluid scale — the app is a desktop-first admin tool and one well-tested cutoff is easier to maintain; tables scroll horizontally rather than reflowing into cards — keeps the mono/tabular-nums density rules intact and required no per-screen template redesign.

## Testing

- `npm run lint` — clean.
- `npm test` (Karma, ChromeHeadlessNoSandbox) — 141/141 green, including 4 new `ShellComponent` specs (hamburger toggle + backdrop rendering, backdrop close, nav-item close, Escape close).
- `npm run build` — production build succeeds.
- Manual verification with Playwright against `ng serve` at 375×812: dashboard, matches, staffer, match form drawer, and the open nav render correctly; `scrollWidth - clientWidth` is 0px (no page-level horizontal overflow) on /dashboard, /matches, /referees, /standings, /configuration and /import; desktop at 1400×900 is visually unchanged. Backend was not running, so screens showed empty states + the expected error toast — layout verification was unaffected.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01FZ1Eodm9ufdbKzQZRwcq5e)_


src/main/webapp/src/app/component/common/icon/icon.component.ts
```diff
@@ -29,6 +29,9 @@ const ICON_PATHS = {
   whistle:    'M4 12a6 6 0 1 0 8.5-5.5L20 4l-1 6h-3',
   grid:       'M4 4h7v7H4zM13 4h7v7h-7zM4 13h7v7H4zM13 13h7v7h-7z',
   list:       'M4 6h16M4 12h16M4 18h16',
+  // Same three-line glyph as `list` (lucide draws them identically) — separate name so
+  // the mobile nav toggle reads as "menu" at the call site.
+  menu:       'M4 6h16M4 12h16M4 18h16',
```
UnLow1: Deliberate duplication: `menu` and `list` share the same lucide path (the glyphs are identical), but the hamburger call site reading `name="menu"` felt worth one duplicated string. If that seems wasteful, the alternative is `menu: ICON_PATHS.list`-style aliasing — not possible with the current `as const` object literal without restructuring, hence the copy.

---

src/main/webapp/src/styles.scss
```diff
@@ -512,6 +517,94 @@ textarea.input { height: auto; min-height: 64px; padding: 8px 10px; line-height:
...
+  // Sidenav becomes an off-canvas panel under the 56px topbar, toggled by the
+  // hamburger in ShellComponent. Kept below .drawer (50) and .modal-backdrop (60)
+  // so overlays opened from a screen still stack above the nav.
+  .sidenav {
+    position: fixed;
+    top: 56px;
+    bottom: 0;
+    left: 0;
+    width: 260px;
+    z-index: 45;
```
UnLow1, line 553: z-index layering rationale: the app's existing overlay stack is `.drawer-backdrop` 40 → `.drawer` 50 → `.modal-backdrop` 60 → `.toast` 70. The mobile nav slots in at 44 (backdrop) / 45 (panel) — above page content, below every overlay a screen can open. In practice the nav and a drawer are never open at the same time (opening anything from a screen implies the nav was already closed), but if that ever happens the drawer correctly wins.

---

src/main/webapp/src/app/component/common/shell/shell.component.spec.ts
```diff
@@ -94,4 +94,63 @@ describe('ShellComponent', () => {
...
+      fixture.detectChanges();
+
+      expect(sidenav().classList).not.toContain('sidenav--open');
+    });
+
+    it('closes the nav when a nav item is clicked', () => {
+      hamburger().click();
+      fixture.detectChanges();
+
+      // ctrlKey makes RouterLink skip the actual navigation (browser would open a new
```
UnLow1, line 137: Test trick worth flagging: a plain `.click()` on the anchor would make `RouterLink` call `router.navigateByUrl('/dashboard')`, which rejects under the spec's empty `provideRouter([])` config and can surface as an unhandled rejection. `RouterLink` intentionally ignores modified clicks (ctrl/cmd/middle-button — the browser would open a new tab), while our `(click)="closeNav()"` binding still fires. If we ever add real route stubs to this spec, this can become a plain `.click()`.

---

src/main/webapp/src/app/component/match-detail/match-detail.component.scss
```diff
@@ -184,3 +184,23 @@
...
   background: var(--accent-tint);
 }
+
+/* ——— <768px (RS-85): hero stacks home/score/away vertically, body panels stack ——— */
+@media (max-width: 767px) {
+  .hero-grid {
+    grid-template-columns: 1fr;
+    padding: 20px 16px;
+    gap: 16px;
+    text-align: center;
```
UnLow1, line 194: Design decision: on desktop the hero is `home | score | away` with the home side right-aligned toward the centered score. Stacked vertically that right-alignment looks like a bug (home name hugging the right edge), so everything centers on mobile — reading order becomes home team → score → away team, matching how fixtures are usually written. Shout if you'd rather keep left-alignment for the stacked variant.

---

src/main/webapp/src/styles.scss
```diff
@@ -512,6 +517,94 @@ textarea.input { height: auto; min-height: 64px; padding: 8px 10px; line-height:
...
+
+  // Page headers: let the action row (search, buttons) wrap under the title.
+  .page-head {
+    flex-wrap: wrap;
+  }
+  .page-head__actions {
+    flex-wrap: wrap;
+  }
+  .page-head__actions .input {
+    flex: 1 1 160px;
```
UnLow1, line 582: The `1 1 160px` basis makes the search input greedy on its own wrapped row (it grows to the full width) while the Filter / New-buttons stay intrinsic and wrap onto the next line when needed — that's the layout visible in the 375px screenshots. Scoped to `.page-head__actions .input` so inputs inside panels/forms are unaffected.

---

UnLow1 (commented):
Self-review notes — context on the discussable spots in this diff (icon duplication, z-index layering, the spec's modified-click trick, mobile hero alignment, search-input flex).

---

src/main/webapp/src/app/component/common/icon/icon.component.ts
UnLow1: Applied in 0dc2af0 — the path now lives in a shared `THREE_LINES` const used by both `list` and `menu`, so the glyphs can't drift apart. Lint + all 141 tests still green.

---

src/main/webapp/src/styles.scss
UnLow1: No change needed — the 44/45 slot is intentional so any drawer or modal opened from a screen always stacks above the nav; keeping it below 50 also means we never have to coordinate Escape handling between the nav and the drawer.

---

src/main/webapp/src/app/component/common/shell/shell.component.spec.ts
UnLow1: Keeping as-is — adding route stubs to this spec just to allow a plain `.click()` would couple the shell spec to the app's route table for no assertion benefit. The inline comment documents the trick for the next reader.

---

src/main/webapp/src/app/component/match-detail/match-detail.component.scss
UnLow1: Keeping the centered variant — it matches the score's existing centering, so the stacked hero reads as one column instead of three mixed alignments. Easy to revisit in a follow-up if it looks off with real long team names.

---

src/main/webapp/src/styles.scss
UnLow1: No change needed — verified in the 375px screenshots: the search input takes its own full-width row and the buttons wrap below, which is the intended mobile layout. Desktop is unaffected since the rule lives inside the media query.

---

PR #95, opened 2026-07-17
RS-89: Add component specs for the remaining redesigned screens

UnLow1:
## Ticket

[RS-89](https://referee-staffer.youtrack.cloud/issue/RS-89) — continuation of RS-81 (PR #74). Five redesigned screens were still shipping without any component tests: **dashboard**, **match-detail**, **referee-profile**, **configuration** and **grade-list**. All of them carry non-trivial derived logic on signals (queue selection, client-side joins, server-value-vs-local-fallback stats, config grouping/editing), so regressions there would go unnoticed by the existing suite.

## Plan

1. Read each component and map out its testable logic: computed signals, service joins in `ngOnInit`, route handling, and user actions.
2. Write one spec per screen following the RS-81 conventions: TestBed with services as `jasmine.createSpyObj` returning `of(...)`, `ActivatedRoute` as a minimal snapshot, `provideRouter([])` for templates that render `routerLink`, and `spyOn(router, 'navigate')` for redirect assertions.
3. Coverage targets per screen:
   - *dashboard* — upcoming-queue selection and difficulty sum, top-referees ranking with the experience fallback and the 7-item cap, pool-wide average grade, standings zones and bar scaling;
   - *match-detail* — id-from-route join (teams/standings/assigned referee), redirect on a missing id, backend breakdown flags vs. the local derivation, candidate ranking, `assign()` via the update endpoint, compare rows, score line, mirrored bars;
   - *referee-profile* — history filtered to the routed referee, lookups only for referenced ids (and skipped when there are none), server-enriched stats vs. local fallbacks, top-cities aggregation, hero labels, edit deep-link;
   - *configuration* — backend-driven grouping with the unknown-group fallback, value parsing and the NaN guard, reset baseline, save round-trip (payload, `saving` flag, refreshed adoption, error path);
   - *grade-list* — client-side row join with unresolved-grade filtering and queue-desc sort, skipped lookups, short-code fallback, meter/kind scaling, edit routing, delete detaching the grade from its match.
4. Verify with `npm ci`, `npm run lint` and a headless `ng test` run.

## Changes

Five new spec files, no production code touched:

- `dashboard.component.spec.ts` (8 tests)
- `match-detail.component.spec.ts` (10 tests)
- `referee-profile.component.spec.ts` (11 tests)
- `configuration.component.spec.ts` (7 tests)
- `grade-list.component.spec.ts` (6 tests)

Design decisions:

- **Sync-first stubbing** — services return `of(...)`, so the nested `forkJoin`s resolve synchronously and most tests need no async plumbing. The one exception is the empty-id branch in `referee-profile` / `grade-list`, where the components fall back to `Promise.resolve([])` inside `forkJoin`; those two tests use `fakeAsync`/`tick()` to flush the microtask.
- **Real router for `routerLink` templates** — dashboard, match-detail and referee-profile render `routerLink`, so their specs use `provideRouter([])` (same as `shell.component.spec.ts`) with a `spyOn(router, 'navigate')` for the redirect/edit assertions, instead of a bare `Router` spy object.
- **`TestBed.resetTestingModule()` inside `create()`** for match-detail and referee-profile — several tests re-create the component with different stubs (e.g. enriched vs. un-enriched referee), and `configureTestingModule` cannot be called twice on an instantiated module.
- **Contradictory breakdown fixture** in match-detail — the backend flags deliberately disagree with what the local derivation would compute, proving the backend wins once loaded and the fallback only covers the loading window.

## Testing

- `npm run lint` — clean.
- `CHROME_BIN=<chromium> npx ng test --browsers=ChromeHeadlessNoSandbox --watch=false` — **179/179 SUCCESS** (42 of them new).
- Backend untouched, so `maven.yml` is not triggered; only `frontend.yml` applies to this diff.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01CesqpHtoGTrru4gBHrM2jH)_


src/main/webapp/src/app/component/match-detail/match-detail.component.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+  it('prefers the backend breakdown flags over the local derivation', async () => {
+    const component = (await create(11)).componentInstance;
+
+    expect(component.sameCity()).toBeFalse();
+    expect(component.isTopMatch()).toBeFalse();
+    expect(component.isRelegationMatch()).toBeTrue();
+  });
+
+  it('falls back to locally derived flags while the breakdown is absent', async () => {
+    matchService.getDifficultyBreakdown.and.returnValue(of(null as unknown as DifficultyBreakdown));
```
UnLow1, line 136: The `null as unknown as DifficultyBreakdown` cast is deliberately ugly: `MatchService.getDifficultyBreakdown` is typed as always emitting a breakdown, but the component's computed flags explicitly handle the "not loaded yet" window via `this.breakdown()?.flags`. Emitting `null` is the cheapest way to pin that fallback path without introducing a timed observable. If the service type ever becomes `DifficultyBreakdown | null`, this cast should be dropped.

---

src/main/webapp/src/app/component/referee-profile/referee-profile.component.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    expect(navigateSpy).toHaveBeenCalledWith(['/referees']);
+    expect(refereeService.findById).not.toHaveBeenCalled();
+  });
+
+  it('skips the team and grade lookups for a referee with no matches', fakeAsync(() => {
+    matchService.findAll.and.returnValue(of([makeMatch(13, {refereeId: 999})]));
+
+    let component!: RefereeProfileComponent;
+    create(7).then(fixture => component = fixture.componentInstance);
```
UnLow1, line 124: This `.then(...)` + `tick()` shape (instead of `await`) is intentional: the test body must stay synchronous inside `fakeAsync`, and `tick()` both resolves the `create()` promise chain and flushes the `Promise.resolve([])` fallbacks the component feeds into `forkJoin` when a referee has no matches. An `async` test with `await create(7)` would leave those microtasks unflushed at assertion time in a less controlled way. Same pattern as the analogous test in `grade-list.component.spec.ts`.

---

src/main/webapp/src/app/component/configuration/configuration.component.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    component.setValue('GRADE_MULTIPLIER', 62.5);
+    component.setValue('DERBY_INCREMENT', 11);
+    component.reset();
+
+    expect(component.valueFor('GRADE_MULTIPLIER')).toBe(50);
+    expect(component.valueFor('DERBY_INCREMENT')).toBe(10);
+  });
+
+  it('saves the edited values and adopts the refreshed response', () => {
+    const response = new Subject<Config[]>();
```
UnLow1, line 78: A `Subject` instead of `of(...)` here so the test can observe `saving() === true` while the request is in flight — with a synchronous `of(...)` the flag would flip back to `false` before the first assertion. This mirrors the `Subject`-for-timing convention noted in the RS-81 specs (e.g. staffer).

---

src/main/webapp/src/app/component/dashboard/dashboard.component.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+      awayScore: undefined,
+      ...overrides
+    } as Match;
+  }
+
+  const matches: Match[] = [
+    // Queue 1 fully played — must not count as upcoming.
+    makeMatch(11, {queue: 1, homeScore: 2, awayScore: 0, hardnessLvl: 50}),
+    // Queue 2 has unplayed fixtures → it is the upcoming queue.
+    makeMatch(12, {queue: 2, hardnessLvl: 80.2}),
```
UnLow1, line 41: The fractional hardness values (80.2 + 120.4) are chosen on purpose: they sum to 200.6, so the `upcomingDifficulty()` expectation of `201` actually proves the `Math.round` on the aggregate — integer fixtures would pass even if the rounding was dropped.

---

UnLow1 (commented):
Self-review notes on the discussable spots in this diff — context for each below.

---

src/main/webapp/src/app/component/match-detail/match-detail.component.spec.ts
UnLow1: Reviewed — keeping the cast. The alternative (a fixture breakdown with `flags: undefined`) would need the exact same kind of cast while being less honest about what the component actually receives during the loading window. Revisit only if the service signature ever widens to `DifficultyBreakdown | null`.

---

src/main/webapp/src/app/component/referee-profile/referee-profile.component.spec.ts
UnLow1: Reviewed — keeping as is. The pattern is verbose but it keeps microtask flushing under `tick()`'s control, and it stays symmetric with the grade-list spec so the next reader sees one convention for the `Promise.resolve` fallback branch, not two.

---

src/main/webapp/src/app/component/configuration/configuration.component.spec.ts
UnLow1: Reviewed — keeping the `Subject`. It's the only way to assert the in-flight `saving()` state, and the separate "clears the saved marker" test uses plain `of(...)` where timing doesn't matter, so the extra machinery stays confined to the one test that needs it.

---

src/main/webapp/src/app/component/dashboard/dashboard.component.spec.ts
UnLow1: Reviewed — no change needed; the comment is informational so future edits to these fixtures don't accidentally trade the fractional values for round numbers and silently weaken the rounding assertion.

---

PR #96, opened 2026-07-19
RS-8: Generate referee assignment PDF for a queue

UnLow1:
# Ticket

[RS-8 — Generate PDF with referee assignment to given queue](https://referee-staffer.youtrack.cloud/issue/RS-8)

The staffer produces a cast for a queue, but there was no way to hand it out — the assignments lived only on the Staffer screen. This adds a downloadable PDF assignment sheet for a given queue: the printable/mailable counterpart of the cast table, and the prerequisite for RS-9 (emailing the sheet to referees).

# Plan

1. Pick a PDF library: **OpenPDF 3.0.5** (`com.github.librepdf:openpdf`, `org.openpdf.*` packages) — actively maintained LGPL/MPL fork of iText 2, no transitive baggage, works on Java 25.
2. Backend rendering: new `AssignmentPdfService` that loads all matches of a queue sorted by date (new `MatchRepository.findAllByQueueOrderByDateAsc`) and renders an A4 sheet — title, generation timestamp, table of date / home / away / referee. Empty queue → 404 via a new queue-flavoured `MatchNotFoundException` constructor, reusing the existing `RestExceptionHandler` mapping.
3. Endpoint: `GET /api/matches/queue/{queue}/pdf` on `MatchController`, `application/pdf` with a `Content-Disposition: attachment` filename.
4. Frontend: `MatchService.downloadAssignmentsPdf` (blob GET) + an **Export PDF** button in the Staffer header that triggers a browser download for the selected queue.
5. Tests: Spock specs for the service (real PDF rendered, content asserted via `PdfTextExtractor`) and the controller slice (headers, body, 404), Jasmine specs for the component flow.

# Changes

- `pom.xml` — OpenPDF 3.0.5 dependency.
- `AssignmentPdfService` — renders the sheet. Design decisions:
  - **Cp1250 encoding** on built-in Helvetica: team/referee names imported from the CSV carry Polish diacritics (`Sędzia`, `Wisła`); the default Cp1252 would mangle them. Standard-14 fonts keep the PDF tiny (~2 KB) vs embedding a TTF.
  - Renders **persisted** assignments — the sheet reflects the last saved cast, not unsaved UI edits. Matches with no referee render as an italic `unassigned`.
  - Date format `dd.MM.yyyy HH:mm` mirrors the CSV importer's format.
- `MatchController.getAssignmentsPdf` — `GET /api/matches/queue/{queue}/pdf`, attachment filename `referee-assignments-queue-{queue}.pdf`; empty queue responds 404 ProblemDetail (surfaced by the frontend's global error toast).
- `MatchNotFoundException` — new `(short queue)` constructor with a `QUEUE_EMPTY` message.
- `MatchRepository` — `findAllByQueueOrderByDateAsc`.
- Staffer screen — **Export PDF** button (download icon, `Exporting…` busy state) in the header actions, next to Generate cast; downloads via a temporary object-URL anchor. Placed in the header (not the cast panel footer) so a previously saved cast can be exported without regenerating.
- `MatchService` (frontend) — blob GET for the endpoint.

# Testing

- **Backend** (`mvn -DskipFrontend=true test`): 134 tests green. `AssignmentPdfServiceSpec` renders a real PDF and asserts extracted text (title, teams with Polish diacritics, dates, referee name, `unassigned` fallback) plus the empty-queue 404 exception; `MatchControllerSpec` covers content type, `Content-Disposition`, body bytes and the 404 ProblemDetail.
- **Frontend** (`npm run lint`, `npm test` headless): lint clean, 140 tests green. New specs cover the download flow (service called with the selected queue, anchor download name, object-URL revocation, busy flag lifecycle, error path).
- Note: the sandbox has no JDK 25 preinstalled — backend tests ran on `openjdk-25-jdk-headless` installed via apt (same major as CI's Temurin 25).

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01DYeb9CdB1cJCuazMiVj7cL)_

### DIFF
--- a/src/main/java/com/jamex/refereestaffer/controller/MatchController.java
+++ b/src/main/java/com/jamex/refereestaffer/controller/MatchController.java
@@ -5,9 +5,14 @@ import com.jamex.refereestaffer.model.dto.DifficultyBreakdownDto;
 import com.jamex.refereestaffer.model.dto.MatchDto;
 import com.jamex.refereestaffer.model.exception.MatchNotFoundException;
 import com.jamex.refereestaffer.repository.MatchRepository;
+import com.jamex.refereestaffer.service.AssignmentPdfService;
 import com.jamex.refereestaffer.service.MatchService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.ContentDisposition;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
@@ -29,11 +34,14 @@ public class MatchController {
     private final MatchRepository matchRepository;
     private final MatchConverter matchConverter;
     private final MatchService matchService;
+    private final AssignmentPdfService assignmentPdfService;
 
-    public MatchController(MatchRepository matchRepository, MatchConverter matchConverter, MatchService matchService) {
+    public MatchController(MatchRepository matchRepository, MatchConverter matchConverter, MatchService matchService,
+                           AssignmentPdfService assignmentPdfService) {
         this.matchRepository = matchRepository;
         this.matchConverter = matchConverter;
         this.matchService = matchService;
+        this.assignmentPdfService = assignmentPdfService;
     }
 
     @GetMapping
@@ -59,6 +67,23 @@ public class MatchController {
     public DifficultyBreakdownDto getMatchDifficulty(@PathVariable Long id) {
         log.info("Computing difficulty breakdown for match {}", id);
         return matchService.computeDifficultyBreakdown(id);
+    }
+
+    /**
+     * Assignment sheet for a queue as a downloadable PDF (RS-8). Renders persisted
+     * assignments — save the cast on the Staffer screen first for it to show up here.
+     */
+    @GetMapping("/queue/{queue}/pdf")
+    public ResponseEntity<byte[]> getAssignmentsPdf(@PathVariable short queue) {
+        log.info("Generating assignment PDF for queue {}", queue);
+        var pdf = assignmentPdfService.generateAssignmentsPdf(queue);
+        var contentDisposition = ContentDisposition.attachment()
+                .filename("referee-assignments-queue-" + queue + ".pdf")
+                .build();
+        return ResponseEntity.ok()
+                .contentType(MediaType.APPLICATION_PDF)
+                .header(HttpHeaders.CONTENT_DISPOSITION, contentDisposition.toString())
+                .body(pdf);
     }
 
     @PostMapping
--- a/src/main/java/com/jamex/refereestaffer/model/exception/MatchNotFoundException.java
+++ b/src/main/java/com/jamex/refereestaffer/model/exception/MatchNotFoundException.java
@@ -3,8 +3,13 @@ package com.jamex.refereestaffer.model.exception;
 public class MatchNotFoundException extends RuntimeException {
 
     public static final String NOT_FOUND = "Match with id = %d has not been found";
+    public static final String QUEUE_EMPTY = "No matches have been found for queue = %d";
 
     public MatchNotFoundException(Long id) {
         super(String.format(NOT_FOUND, id));
+    }
+
+    public MatchNotFoundException(short queue) {
+        super(String.format(QUEUE_EMPTY, queue));
     }
 }
--- a/src/main/java/com/jamex/refereestaffer/service/AssignmentPdfService.java
+++ b/src/main/java/com/jamex/refereestaffer/service/AssignmentPdfService.java
@@ -1,0 +1,126 @@
+package com.jamex.refereestaffer.service;
+
+import com.jamex.refereestaffer.model.entity.Match;
+import com.jamex.refereestaffer.model.entity.Team;
+import com.jamex.refereestaffer.model.exception.MatchNotFoundException;
+import com.jamex.refereestaffer.repository.MatchRepository;
+import org.openpdf.text.Document;
+import org.openpdf.text.DocumentException;
+import org.openpdf.text.Element;
+import org.openpdf.text.Font;
+import org.openpdf.text.PageSize;
+import org.openpdf.text.Paragraph;
+import org.openpdf.text.Phrase;
+import org.openpdf.text.pdf.BaseFont;
+import org.openpdf.text.pdf.PdfPCell;
+import org.openpdf.text.pdf.PdfPTable;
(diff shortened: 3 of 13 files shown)


src/main/java/com/jamex/refereestaffer/model/exception/MatchNotFoundException.java
```diff
@@ -3,8 +3,13 @@
 public class MatchNotFoundException extends RuntimeException {
 
     public static final String NOT_FOUND = "Match with id = %d has not been found";
+    public static final String QUEUE_EMPTY = "No matches have been found for queue = %d";
 
     public MatchNotFoundException(Long id) {
         super(String.format(NOT_FOUND, id));
     }
+
+    public MatchNotFoundException(short queue) {
```
UnLow1, line 12: Discussable: this overload is distinguished from `(Long id)` only by the parameter type (`short` vs `Long`), which is subtle — `new MatchNotFoundException(5L)` and `new MatchNotFoundException((short) 5)` mean different things. It keeps the existing 404 mapping in `RestExceptionHandler` for free, which is why I chose it over a dedicated `EmptyQueueException`. If a third "not found" flavour ever appears, a static factory (`forQueue(short)`) or a separate exception would be the cleaner refactor.

---

src/main/java/com/jamex/refereestaffer/service/AssignmentPdfService.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        log.info("Rendering assignment PDF for queue {} with {} matches", queue, matches.size());
+        return render(queue, matches);
+    }
+
+    private byte[] render(short queue, List<Match> matches) {
+        try {
+            // Cp1250 (Central European) instead of the default Cp1252 — team and referee
+            // names imported from the CSV carry Polish diacritics.
+            var baseFont = BaseFont.createFont(BaseFont.HELVETICA, BaseFont.CP1250, BaseFont.NOT_EMBEDDED);
```
UnLow1, line 65: Font tradeoff worth knowing about: `NOT_EMBEDDED` + standard-14 Helvetica means the viewer substitutes its own font, and Cp1250 covers exactly the Central European set (all Polish diacritics included — the spec asserts `Sędzia`/`Wisła` render). If names outside Cp1250 ever show up (e.g. a referee with a Czech `ů` is fine, but Cyrillic is not), the fix is to embed a Unicode TTF (`BaseFont.IDENTITY_H` + e.g. Liberation Sans from `openpdf-fonts-extra`) at the cost of ~200 KB per PDF. Not needed for the current data.

---

src/main/java/com/jamex/refereestaffer/controller/MatchController.java
```diff
@@ -61,6 +69,23 @@ public DifficultyBreakdownDto getMatchDifficulty(@PathVariable Long id) {
         return matchService.computeDifficultyBreakdown(id);
     }
 
+    /**
+     * Assignment sheet for a queue as a downloadable PDF (RS-8). Renders persisted
+     * assignments — save the cast on the Staffer screen first for it to show up here.
+     */
+    @GetMapping("/queue/{queue}/pdf")
```
UnLow1, line 76: Placement decision: this lives on `MatchController` (`/api/matches/queue/{queue}/pdf`) rather than `StafferController` because it's a read of persisted match data — no staffing logic runs. `StafferController` is deliberately mutation-only (its POST persists a new cast). The `/queue/{queue}` path segment also leaves room for a future JSON `GET /api/matches/queue/{queue}` if the frontend ever needs matches-by-queue without staffing.

---

src/main/webapp/src/app/component/staffer/staffer.component.ts
```diff
@@ -139,6 +140,29 @@ export class StafferComponent {
...
+        anchor.href = url;
+        anchor.download = `referee-assignments-queue-${this.queue()}.pdf`;
+        // Firefox needs the anchor in the DOM for a programmatic download click.
+        document.body.appendChild(anchor);
+        anchor.click();
+        anchor.remove();
+        URL.revokeObjectURL(url);
+        this.exporting.set(false);
+      },
+      error: () => this.exporting.set(false)
```
UnLow1, line 162: Known rough edge: on a 404 (queue with no matches) the global `httpErrorInterceptor` shows the generic `Request failed (HTTP 404)` toast instead of the backend's `No matches have been found for queue = N`. The response body is a `Blob` here (because of `responseType: 'blob'`), so `error.error?.detail` is undefined — reading it would need an async `Blob.text()` + JSON parse in the interceptor. Left out deliberately to keep the interceptor synchronous; can be added if the generic toast proves confusing.

---

UnLow1 (commented):
Self-review notes — context for the discussable spots in this diff (exception overload, font encoding, endpoint placement, blob error toast).

---

src/main/webapp/src/app/component/staffer/staffer.component.ts
UnLow1: Fixed in 589b674 — the interceptor now reads JSON-typed error blobs asynchronously (`Blob.text()` + JSON parse) and toasts the ProblemDetail `detail`, so an empty queue shows "No matches have been found for queue = N" instead of the generic message. Unparseable bodies fall back to the generic toast. Covered by a new `http-error.interceptor.spec.ts` (JSON, blob, broken-blob, offline and rethrow paths); lint clean, 146 frontend tests green.

---

src/main/java/com/jamex/refereestaffer/model/exception/MatchNotFoundException.java
UnLow1: Keeping as is: both call sites are explicit (`Long` boxed id in controllers/services, `short queue` only in `AssignmentPdfService`), and the alternative — a new exception class plus a handler registration — buys no behavioural difference for a single caller. The static-factory refactor is noted for when a third variant appears.

---

src/main/java/com/jamex/refereestaffer/service/AssignmentPdfService.java
UnLow1: Keeping the non-embedded standard font: the imported data is Polish-league only, Cp1250 covers it fully (asserted in `AssignmentPdfServiceSpec` with `Sędzia`/`Wisła`), and the resulting PDFs stay ~2 KB — which matters for RS-9 (emailing sheets to referees). Embedding a Unicode TTF is the documented escape hatch if the data ever outgrows Cp1250.

---

src/main/java/com/jamex/refereestaffer/controller/MatchController.java
UnLow1: Keeping the placement: exporting must work for a cast saved in an earlier session without re-running the staffer (which would mutate assignments), so the endpoint belongs with match reads, not staffing. If more per-queue reads appear later, extracting a `QueueController` is the natural follow-up.

---

PR #98, opened 2026-07-20
RS-95: Add specs for ToastService, UiSettingsService and the logic-bearing atoms

UnLow1:
## Ticket

[RS-95](https://referee-staffer.youtrack.cloud/issue/RS-95) — Testy frontendu: ToastService, UiSettingsService i atomy z logiką (team-pill, ref-avatar)

This closes the remaining frontend unit-test gap left after RS-80 (HTTP services), RS-81 (hot spots, forms, overlays, lists) and RS-89 (remaining screens): the two non-HTTP services and the two atoms that contain real logic had neither a spec nor a ticket covering them.

## Plan

1. **`ToastService`** — spec with `fakeAsync` + `tick` (no real timers): default kind `info`, explicit `error` kind, auto-hide exactly at `HIDE_AFTER_MS`, and the replacement behaviour — a second `show()` overwrites the current toast **and restarts** the hide timer.
2. **`UiSettingsService`** — the constructor reads localStorage, so each test arranges the persisted state first and injects the service inside the test body; cover the clean-slate defaults, restore-on-startup for all three keys (`theme`, `admin.visible`, `staffer.explainer`), each toggle's signal flip + localStorage write, and the `data-theme` stamp on `<html>`. localStorage and the `data-theme` attribute are reset around every test.
3. **`TeamPillComponent`** — backend `short` code rendering, the 3-letter uppercase fallback derived from `name`, the empty no-team state, and the home/away colour variants (`team-pill__short--home`).
4. **`RefAvatarComponent`** — `initials()`/`fullName()` rendering, `sm` vs `lg` email visibility (`avatar--lg` class), missing email in `lg`, and partial-name edge cases.

The optional low-priority smoke tests for the purely presentational atoms (`icon`, `chip`, `panel`, `kpi`, `meter`, `seg`, toast component) are deliberately left out, as the ticket marks them optional.

## Changes

Four new spec files, no production code touched:

- `src/app/service/toast.service.spec.ts` — 5 tests
- `src/app/service/ui-settings.service.spec.ts` — 5 tests; a shared `reset()` runs in both `beforeEach` **and** `afterEach` so the localStorage keys and the `data-theme` attribute the toggles write cannot leak into whatever suite runs next
- `src/app/component/common/team-pill/team-pill.component.spec.ts` — 5 tests
- `src/app/component/common/ref-avatar/ref-avatar.component.spec.ts` — 6 tests

Conventions follow RS-81/RS-89: standalone-component TestBed setup, `fixture.componentRef.setInput` for signal inputs, assertions against rendered DOM rather than component internals where a template exists.

## Testing

- `npm run lint` — clean.
- `npm test -- --watch=false --browsers=ChromeHeadlessNoSandbox --no-progress` — **269/269 SUCCESS** (21 new tests on top of the previous 248), same invocation as `frontend.yml`.
- Environment note: the container ships Node v22.22.2, one patch below Angular CLI 22's minimum (≥22.22.3), so lint/tests were run locally with Node 24.15.0 — the exact version pinned in `pom.xml` and used by CI.

🤖 Generated with Claude Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01HpttWxouks7uBBfJ2qaDRu)_

### DIFF
--- a/src/main/webapp/src/app/component/common/team-pill/team-pill.component.spec.ts
+++ b/src/main/webapp/src/app/component/common/team-pill/team-pill.component.spec.ts
@@ -1,0 +1,59 @@
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {TeamPillComponent} from './team-pill.component';
+import {Team} from '../../../model/team';
+
+describe('TeamPillComponent', () => {
+  const team: Team = {id: 1, name: 'Legia Warszawa', city: 'Warszawa', points: 30, short: 'LEG'};
+
+  let fixture: ComponentFixture<TeamPillComponent>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({imports: [TeamPillComponent]}).compileComponents();
+    fixture = TestBed.createComponent(TeamPillComponent);
+  });
+
+  function el(): HTMLElement {
+    return fixture.nativeElement as HTMLElement;
+  }
+
+  function shortEl(): HTMLElement {
+    return el().querySelector('.team-pill__short') as HTMLElement;
+  }
+
+  it('renders the backend short code and the full team name', () => {
+    fixture.componentRef.setInput('team', team);
+    fixture.detectChanges();
+
+    expect(shortEl().textContent?.trim()).toBe('LEG');
+    expect(el().querySelector('.team-pill__name')?.textContent).toContain('Legia Warszawa');
+  });
+
+  it('derives an uppercase 3-letter code from the name when short is absent', () => {
+    fixture.componentRef.setInput('team', {...team, short: undefined});
+    fixture.detectChanges();
+
+    expect(shortEl().textContent?.trim()).toBe('LEG');
+  });
+
+  it('renders empty without a team', () => {
+    fixture.detectChanges();
+
+    expect(shortEl().textContent?.trim()).toBe('');
+    expect(el().querySelector('.team-pill__name')?.textContent?.trim()).toBe('');
+  });
+
+  it('marks the home side with the dark variant by default', () => {
+    fixture.componentRef.setInput('team', team);
+    fixture.detectChanges();
+
+    expect(shortEl().classList).toContain('team-pill__short--home');
+  });
+
+  it('drops the dark variant for the away side', () => {
+    fixture.componentRef.setInput('team', team);
+    fixture.componentRef.setInput('side', 'away');
+    fixture.detectChanges();
+
+    expect(shortEl().classList).not.toContain('team-pill__short--home');
+  });
+});
--- a/src/main/webapp/src/app/component/common/team-pill/team-pill.component.spec.ts
+++ b/src/main/webapp/src/app/component/common/team-pill/team-pill.component.spec.ts
@@ -3,7 +3,9 @@ import {TeamPillComponent} from './team-pill.component';
 import {Team} from '../../../model/team';
 
 describe('TeamPillComponent', () => {
-  const team: Team = {id: 1, name: 'Legia Warszawa', city: 'Warszawa', points: 30, short: 'LEG'};
+  // short deliberately differs from the name-derived prefix ("LEG") so the tests can
+  // tell the backend code apart from the fallback.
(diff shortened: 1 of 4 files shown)


src/main/webapp/src/app/component/common/team-pill/team-pill.component.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  function shortEl(): HTMLElement {
+    return el().querySelector('.team-pill__short') as HTMLElement;
+  }
+
+  it('renders the backend short code and the full team name', () => {
+    fixture.componentRef.setInput('team', team);
+    fixture.detectChanges();
+
+    expect(shortEl().textContent?.trim()).toBe('LEG');
```
UnLow1: Weak assertion: the fixture's `short: 'LEG'` happens to equal what the fallback would derive from `'Legia Warszawa'.slice(0, 3)`, so this test would still pass if the component ignored `Team.short` entirely and always derived from the name. The fixture's short code should differ from the name-derived prefix to actually prove precedence.

---

src/main/webapp/src/app/service/toast.service.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  it('starts with no toast', () => {
+    expect(service.toast()).toBeNull();
+  });
+
+  it('shows a toast with the default info kind', fakeAsync(() => {
+    service.show('Import finished');
+
+    expect(service.toast()).toEqual({message: 'Import finished', kind: 'info'});
+    tick(HIDE_AFTER_MS);
```
UnLow1, line 21: Context for the reader: this trailing `tick(HIDE_AFTER_MS)` (also in the error-kind test below) isn't asserting anything — `show()` always schedules a hide timer, and `fakeAsync` fails a test that ends with a timer still queued. Flushing it here keeps the test focused on the "shows" behaviour without a `discardPeriodicTasks`-style workaround.

---

src/main/webapp/src/app/service/ui-settings.service.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  // afterEach too: the toggles write localStorage and stamp <html>, and that state
+  // must not leak into whatever suite runs next.
+  beforeEach(reset);
+  afterEach(reset);
+
+  // The constructor reads localStorage, so each test injects the service itself
+  // after arranging the persisted state.
+  function inject(): UiSettingsService {
+    return TestBed.inject(UiSettingsService);
```
UnLow1, line 20: Deliberate deviation from the usual "inject in `beforeEach`" pattern: `UiSettingsService` reads localStorage in its constructor (signal initial values + the `data-theme` stamp), so the service must be created *after* each test has arranged the persisted state. `TestBed.inject` still memoizes within a test, and the TestBed reset between tests guarantees a fresh instance per test.

---

UnLow1 (commented):
Self-review: one real finding (the team-pill precedence assertion) plus two context notes for the reader.

---

src/main/webapp/src/app/component/common/team-pill/team-pill.component.spec.ts
UnLow1: Fixed in bee070a: the fixture now uses `short: 'LGW'`, which differs from the name-derived `'LEG'`, so the first test proves `Team.short` takes precedence and the fallback test still exercises the derivation. Lint + full Karma suite re-run green (269/269).

---

src/main/webapp/src/app/service/toast.service.spec.ts
UnLow1: No change needed — this is a note for future readers, not a defect. The flush is required by `fakeAsync` and intentionally kept out of the test's assertions.

---

src/main/webapp/src/app/service/ui-settings.service.spec.ts
UnLow1: No change needed — the deviation is deliberate and documented in the spec itself; injecting in `beforeEach` would freeze the constructor-read state before tests could arrange localStorage.

---

